PR #1, opened 2015-04-09
Initial commit that contains playbooks and roles for setting up the admin terminal

tanderegg:
It would be great if someone could test this out by cloning the repo and running "vagrant up" as described in the README, thanks!


deploy/Vagrantfile
```diff
@@ -0,0 +1,91 @@
...
+  ### COMMON VARIABLES
+  
+  VAGRANT_ROOT = File.dirname(File.expand_path(__FILE__))
+
+  ### SHARED COMPONENT
+  
+  # GlusterFS File Server 1
+  disk_file_1 = File.join(VAGRANT_ROOT, 'gfs_brick_1.vdi')
+  config.vm.define :file_server_1 do |fs_config|
+    fs_config.vm.box = "chef/centos-6.5"
```
exafox, line 22: Is this notation compatible with older versions of vagrant? Would adding box_url setting make the Vagrantfile more broadly compatible?

deploy/Vagrantfile
```diff
@@ -0,0 +1,91 @@
...
+
+    fs_config.vm.network :private_network, ip: "10.0.1.5"
+  end
+
+  # Admin Terminal VM
+  config.vm.define :admin_terminal do |adterm_config|
+    adterm_config.vm.box = "chef/centos-6.5"
+    adterm_config.vm.host_name = "admin.test.dev"
+
+    adterm_config.vm.provision "ansible" do |ansible|
```
exafox, line 80: Is it worthwhile to more accurately simulate the way this box is provisioned in a production environment? (i.e. Don't use ansible to build the ansible host...)

deploy/roles/python-build/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+    dest: "~/python27"
+    repo: https://github.com/cfpb/python27-for-el6.git
+
+- name: Run bootstrap script
+  sudo: no
+  command: "sh ./bootstrap.sh"
+  args:
+    chdir: "~/python27"
+
+- name: Build python RPM
```
exafox, line 21: I had imagined rpm building living outside the deployment process entirely. Does it make sense to make a bigger deal of the order things happen here and to separate these workloads? (Even if we necessarily run this on vagrant up, perhaps it could live elsewhere to more clearly separate it from provisioning code?)

exafox:
Looking sharp and no errors were encountered when I ran `vagrant up`. There are a couple things to discuss around the sequence of events in bootstrapping the admin terminal. It may not make sense to address any of my comments right now, but lets make sure there is a well established way of thinking about such things?

deploy/deploy_admin_terminal.yml
```diff
@@ -0,0 +1,9 @@
+# deploy_admin_terminal.yml
+---
+- hosts: admin_terminal
+  sudo: yes
+  roles:
+    - iptables
+    - common
+    - {role: python, when: "custom_repo != None"}
```
exafox, line 8: What's it called when you use ansible to install ansible? ...anception?  Should that happen here?

deploy/Vagrantfile
tanderegg, line 22: No, Vagrant doesn't support backwards compatibility like that, you have to choose which configuration notation to use (hence the "2" in Vagrant.configure()).  The vast majority of people will be on Vagrant 1.1+, which uses version 2 notation (the latest version is 1.7.2).  If we really want to, we could add parallel sections for each VM using version 1, but I don't think that's really needed unless someone raises it as an issue.

deploy/Vagrantfile
tanderegg, line 80: As I see it, we'll be provisioning the Admin Terminal in either of 2 ways: from our local machine, or from Jenkins.  If we're doing it from our local machine, we can just install Ansible locally.  In the Jenkins build, we can install Ansible on the Admin Terminal before running the provisioning scripts.  Either way, I think we benefit from being able to reuse the Ansible roles (e.g. easier to maintain, easier to read) , rather than having to rewrite the commands in bash or something.

deploy/roles/python-build/tasks/main.yml
tanderegg, line 21: This was partly for convenience, although I do see having Python 2.7 installed as a dependency of the platform overall.  If we consider it a dependency, then we either have to install it as part of the build or require users to install it manually first (or to provide the RPM manually in some way).  By allowing the build to happen within the provisioning process, we make it easier for the user to use the platform with minimal manual configuration.  There is an option to use a custom repository instead, which is a way of avoiding the build if you want, and I think that's sufficient flexibility.  I'm not sure where else I would put it, that wouldn't just make things more complex unnecessarily.  Open to suggestions though.

deploy/deploy_admin_terminal.yml
tanderegg, line 8: Anception works for me :)  Agreed that should be added in, I'll commit shortly.

exafox:
Merging this. 

I think we will need to sacrifice some Vagrant workflow niceties in the future to accommodate a more production like build/deployment flow, but lets get there gradually. Perhaps we should take this question up at the next code review?

PR #2, opened 2015-04-23
Implementation for the Data Enclave proxy server

tanderegg:
Finally got LDAP working fully so the restricted area is only accessable by those in the "admin" LDAP group.

### DIFF
--- a/deploy/Vagrantfile
+++ b/deploy/Vagrantfile
@@ -89,4 +89,46 @@ Vagrant.configure(2) do |config|
 
     adterm_config.vm.network :private_network, ip: "10.0.1.2"
   end
+
+  # Kerberos Server VM - for testing purposes only
+  config.vm.define :kerberos_server do |kerberos_config|
+    kerberos_config.vm.box = "chef/centos-6.5"
+    kerberos_config.vm.host_name = "kdc.test.dev"
+
+    kerberos_config.vm.provision :shell,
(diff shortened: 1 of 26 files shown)


deploy/Vagrantfile
```diff
@@ -88,4 +89,46 @@ Vagrant.configure(2) do |config|
...
+
+    proxy_config.vm.provision "ansible" do |ansible|
+      ansible.playbook = "deploy_proxy_server.yml"
+
+      ansible.groups = {
+        "proxy_server" => ["proxy_server"],
+        "all_groups:children" => ["proxy_server"]
+      }
+    end
+
```
exafox, line 131: Can we forward the port apache listens on to the host? 

`proxy_config.vm.network "forwarded_port", guest: 80, host: 8080` seems to do the trick just fine for me...

exafox:
Can merge whenever. Builds just fine on my machine

deploy/Vagrantfile
tanderegg, line 131: We can, although currently you can view the proxy server just by going to port 80 on 10.0.1.3 from the host.  Do we want to change ports for another reason?

tanderegg:
Cool, go ahead and do it when you get a chance then, I'll issue a new one
when I've got the iptables stuff done, likely Monday.

deploy/Vagrantfile
exafox, line 131: When your Dev box is a headless server (waves) you don't have a web browser that can see the private network...

PR #3, opened 2015-04-29
IP Tables Configuration

tanderegg:
@exafox and @salimzayat, please take a minute to review the rules here (and in our internal repo) to double check that everything looks right.  Thanks!


salimzayat:
Few thoughts:

1)  I noticed the rules are being inserted at index 3 instead of appended
to the end.  Is there a reason for that?

2) With regards to this chunk at the top:
- name: Temporarily allow all- command: "iptables -P {{ iptables_chain }}
  ACCEPT"+ command: "iptables -P INPUT ACCEPT" - name: Flush existing rules
  command: "iptables -F"

What is the purpose of it?  We accept all incoming before we set them.  Why?

Other than that, I think it is quite clear.

tanderegg:
1) That's because the last rule is to send everything to LOGDROP, so if we inserted rules after that point they would never be reached.  3 is fairly arbitrary, it's just the "slot" that all custom rules will go in in other roles.  The order doesn't matter other than being before the LOGDROP rule (as I understand it at least)

2) That's to prevent iptables from blocking the SSH port during Ansible's execution, which would happen during the FLUSH phase if the INPUT policy was not set to ACCEPT.  That would trigger iptables to cut off Ansible's session.

PR #4, opened 2015-05-14
Develop

salimzayat:
Here is the code for EoD.  Seems to work in Dev.

Sorry, it isn't fully complete; I haven't had a chance to get the passwords into secrets.  I will do that now.


tanderegg:
Use #5 instead.

PR #50, opened 2016-01-05
Update master to reflect current configurations

tanderegg:


tanderegg:
Note to self: merge once Attachment ETL is finalized

PR #70, opened 2016-04-12
Added better logic to support UX for standing up db_server

tanderegg:
@frank0651 - These changes automatically create a .vault_password file (empty), and prompt the user to set the DB_SERVER_EXTRA_VARS environment variable when vagrant up db_server is run.  The PPAS installer is now downloaded directly from EnterpriseDB instead of the requiring the file to be present.

### DIFF
--- a/deploy/Vagrantfile
+++ b/deploy/Vagrantfile
@@ -1,9 +1,16 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
+require 'fileutils'
 
 Vagrant.configure(2) do |config|
 
   ### COMMON VARIABLES
+
+  # First make sure that an .vault_password file exists, and if it doesn't
+  # then create an empty file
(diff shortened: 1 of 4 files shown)


tanderegg:
Sorry, I meant @frank0631, see above comment.

frank0631:
Glad to get PPAS straight from it's http address

frank0631:
I had to comment out the following Ansible task: 

```
 # - name: Install extensions and supplementary packages
 #   yum:
 #     name: "{{ item }}"
 #     state: present
 #   with_items: pp_extensions
```

because I also had commented out all of the postgres extensions in pp_extensions on the yaml file: local_db_settings.yml

tanderegg:
@frank0631 for pp_extensions, just set it to: `pp_extensions: []`, it has to be a list not an empty value.

PR #75, opened 2016-04-25
Spark

olivermel:


deploy/roles/spark/defaults/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+spark_tar_file: "spark-1.5.2-bin-without-hadoop.tgz"
+spark_mirror: "http://apache.claz.org/spark/spark-1.5.2/spark-1.5.2-bin-without-hadoop.tgz"
+spark_src_dir: "/usr/local/src"
+spark_conf_dir: "/etc/spark"
+spark_usr_parent_dir: "/usr/lib"                                                            #spark extraction folder
```
tanderegg: @olivermel Can you put the comments left-aligned and above the setting they refer to?  Having them all the way to the right can make it easy to miss them.

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
+---
+- name: Create service account for Spark
+  user: name={{ spark_user }}
```
tanderegg: @olivermel Sorry to nitpick, but could you use the same style as the other roles in Aurora?  So like this:
user:
  name: "{{ spark_user }}"
  system: yes
  etc...

Thanks!

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+        system=yes
+        home={{ spark_lib_dir }}
+        shell={{ spark_user_shell }}
+        state=present
+        groups="{{ spark_user_groups | join(',') }}"
+  tags: ["spark-user"]
```
tanderegg: How are you using tags in this case?  Before I've set a tag name based on the role name, to make it easy to run a single role when doing development/test, but it seems you have a more complex scheme here.  This might be a good idea, just curious what your plan is.

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,86 @@
...
+        mode=1777
+        state=directory
+  with_items:
+    - "{{ spark_log_dir }}"
+
+- name: Download Spark
+  get_url: url="{{ spark_mirror }}"                                                             
+           dest="{{ spark_src_dir }}"
+
+- name: Extract Spark
```
tanderegg, line 45: Can you add:

mode: a+r

To ensure that the files are all world readable.

deploy/roles/spark/templates/spark-defaults.conf.j2
```diff
@@ -0,0 +1,32 @@
...
+# spark.master                     spark://master:7077
+# spark.eventLog.enabled           true
+# spark.eventLog.dir               hdfs://namenode:8021/directory
+# spark.serializer                 org.apache.spark.serializer.KryoSerializer
+# spark.driver.memory              5g
+# spark.executor.extraJavaOptions  -XX:+PrintGCDetails -Dkey=value -Dnumbers="one two three"
+
+spark.eventLog.enabled          true
+spark.eventLog.dir              file:///{{ spark_log_dir }}
+spark.authenticate              true
```
tanderegg: Can you make spark.authenticate configurable via an Ansible variable as well?

deploy/roles/spark/templates/spark-defaults.conf.j2
```diff
@@ -0,0 +1,32 @@
...
+# Example:
+# spark.master                     spark://master:7077
+# spark.eventLog.enabled           true
+# spark.eventLog.dir               hdfs://namenode:8021/directory
+# spark.serializer                 org.apache.spark.serializer.KryoSerializer
+# spark.driver.memory              5g
+# spark.executor.extraJavaOptions  -XX:+PrintGCDetails -Dkey=value -Dnumbers="one two three"
+
+spark.eventLog.enabled          true
+spark.eventLog.dir              file:///{{ spark_log_dir }}
```
tanderegg: I think there may be an extra "/" in the file url, could confirm this works correctly as is?

deploy/roles/spark/defaults/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+spark_tar_file: "spark-1.5.2-bin-without-hadoop.tgz"
+spark_mirror: "http://apache.claz.org/spark/spark-1.5.2/spark-1.5.2-bin-without-hadoop.tgz"
+spark_src_dir: "/usr/local/src"
+spark_conf_dir: "/etc/spark"
+spark_usr_parent_dir: "/usr/lib"                                                            #spark extraction folder 
+spark_main_dir: "/usr/lib/spark-1.5.2-bin-without-hadoop"                                   #symlink to the extracted/installed spark
+spark_lib_dir: "/var/lib/spark"
+spark_log_dir: "/var/log/spark"
+spark_run_dir: "/run/spark"
```
tanderegg: This should probably be /var/run/spark, no?

deploy/roles/spark/templates/spark-defaults.conf.j2
```diff
@@ -0,0 +1,32 @@
...
+# spark.master                     spark://master:7077
+# spark.eventLog.enabled           true
+# spark.eventLog.dir               hdfs://namenode:8021/directory
+# spark.serializer                 org.apache.spark.serializer.KryoSerializer
+# spark.driver.memory              5g
+# spark.executor.extraJavaOptions  -XX:+PrintGCDetails -Dkey=value -Dnumbers="one two three"
+
+spark.eventLog.enabled          true
+spark.eventLog.dir              file://{{ spark_log_dir }}
+spark.authenticate              true
```
tanderegg: @olivermel This one got skipped: can you change spark.authenticate's value to a variable, with default of "true"?

deploy/roles/spark/defaults/main.yml
tanderegg: @olivermel This one got missed too, can you align the comments to the left, above the code they refer to, so they won't get missed?

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+  file: 
+    path: "{{ spark_conf_dir }}"
+    state: directory
+
+- name: Make sure Spark directories exists
+  file: 
+    path: "{{ item }}" 
+    owner: "{{ spark_user }}"
+    group: "{{ spark_user }}"
+    mode: 0755, a+x
```
tanderegg: @olivermel This should be 0755 only, you can't use both options.

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+  with_items:
+    - "{{ spark_run_dir }}"
+    - "{{ spark_main_dir }}"
+
+- name: Ensure Spark log directory with appropriate permissions
+  file: 
+    path: "{{ item }}" 
+    owner: "{{ spark_user }}"
+    group: "{{ spark_user }}"
+    mode: 1777, a+x
```
tanderegg: Same here, remove the a+x, and let's make it 1755 if that works (or 1775 if necessary)

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
...
+     dest: "{{ item.dest }}"
+     state: link
+  with_items:
+    - { src: "{{ spark_main_dir }}/conf", dest: "{{ spark_conf_dir }}/conf" }
+
+- name: Spark binaries set up
+  template: 
+    src: spark-bin.j2
+    dest: "/usr/bin/{{ item }}"
+    mode: 0755, a+x
```
tanderegg: Same here, remove a+x

deploy/roles/spark/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+spark_tar_file:             "spark-1.5.2-bin-without-hadoop.tgz"
+spark_mirror:               "http://apache.claz.org/spark/spark-1.5.2/spark-1.5.2-bin-without-hadoop.tgz"
```
tanderegg: @olivermel it turns out that we do need the version with Hadoop, otherwise sparks expects you to install it separately.  Use the version with hadoop2.6, I tested it and it works fine.  Thanks!

deploy/roles/spark/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
+---
+spark_tar_file:             "spark-1.3.1-bin-hadoop2.6.tgz"
```
tanderegg: @olivermel This still needs to be the correct version, 1.5.2, this is the link you should use: http://mirror.cc.columbia.edu/pub/software/apache/spark/spark-1.5.2/spark-1.5.2-bin-hadoop2.6.tgz

deploy/roles/spark/templates/spark-defaults.conf.j2
```diff
@@ -0,0 +1,32 @@
...
+# spark.master                     spark://master:7077
+# spark.eventLog.enabled           true
+# spark.eventLog.dir               hdfs://namenode:8021/directory
+# spark.serializer                 org.apache.spark.serializer.KryoSerializer
+# spark.driver.memory              5g
+# spark.executor.extraJavaOptions  -XX:+PrintGCDetails -Dkey=value -Dnumbers="one two three"
+
+spark.eventLog.enabled          true
+spark.eventLog.dir              file://{{ spark_log_dir }}
+spark.authenticate              {{ spark_authenticate_variable }}
```
tanderegg: Can you remove the word "variable" from these two variables?  They are inherently variables, we don't need to indicate that in the name.

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+    home: "{{ spark_lib_dir }}"
+    shell: "{{ spark_user_shell }}"
+    state: present
+    groups: "{{ spark_user_groups | join(',') }}"
+    comment: "Spark-User"
+
+- name: Make sure Spark configuration directory exists
+  file: 
+    path: "{{ spark_conf_dir }}"
+    state: directory
```
tanderegg, line 15: Set mode explicitly to 0755

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+  with_items:
+    - spark-class
+    - spark-shell
+    - spark-sql
+    - spark-submit
+
+- name: Set up SPARK_HOME environment variables
+  template: 
+    src: "spark-env.sh.j2"
+    dest: "{{ spark_main_dir }}/conf/spark-env.sh"
```
tanderegg, line 73: Add in mode: 0644

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+
+- name: Set up spark defaults config file
+  template: 
+    src: spark-defaults.conf.j2
+    dest: "{{ spark_main_dir }}/conf/spark-defaults.conf"
```
tanderegg, line 79: mode: 0644

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+
+- name: Set up spark log4j config
+  template: 
+    src: log4j.properties
+    dest: "{{ spark_main_dir }}/conf/spark.properties"
```
tanderegg, line 85: mode: 0644

deploy/roles/spark/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+
+- name: Ensure Spark environmental variables are set
+  template:
+    src: spark-env.sh.j2
+    dest: /etc/profile.d/spark.sh
+    mode: 0644
+
+- name: reload profile
```
tanderegg: Remove this task, as Ansible commands existing within a separate session, and the next command will already load anything in /etc/profile.d

PR #93, opened 2016-06-13
redis role

olivermel:


tanderegg:
@olivermel I think this is missing all the files, take a look at the "Files Changed" section.  Shouldn't the PR be from your "redis" branch instead of your "master" branch?

PR #95, opened 2016-06-14
Scala and sbt role....

olivermel:


tanderegg:
@olivermel - This PR includes all of the Redis stuff too.  My guess is you ran "git checkout -b scala" while you were still in your "redis" branch.  git checkout will create a new branch based on whatever branch you are currently on, so normally you have to switch back to "develop" first, then do the git checkout.  Can you create a new branch off of develop to keep the changes separate?

PR #100, opened 2016-06-20
new scala role fixed

olivermel:


tanderegg:
@olivermel Thanks, much better :)

tanderegg:
@olivermel - Could you add tags to the roles/scala/tasks/main.yml file?  Basically like this:
- name: Blah blah
  command: ....
  tags:
  - scala

Just add the "tags: " part to each command.  Then, it can be run like this: `TAGS=scala vagrant provision analytics_terminal`, and only the tasks in this role will be executed.  This makes it much easier to test.

Also, could you rebase against the develop branch?  First switch to develop, then "git pull cfpb develop" (or whatever name you gave to the cfpb/aurora remote), then rebase your branch against that.  Thanks!

deploy/roles/scala/defaults/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+
+scala_version:              2.11.8
+scala_url:                  "http://downloads.lightbend.com/scala/{{ scala_version }}/scala-{{ scala_version }}.tgz"
+
+
```
tanderegg: @olivermel Last thing - Can you clean up the newlines throughout this PR?  Use only 1 newline in between various logical sections of code, here you have 3, other places 2.  It makes everything easier to read.  Thanks!

PR #103, opened 2016-07-11
Maven role 

olivermel:


deploy/deploy_app_server.yml
```diff
@@ -17,6 +17,7 @@
     - pycharm
     - nginx
     - nodejs
+    - maven
```
tanderegg, line 20: This should be added to the "analytics_terminal" playbook as well.

tanderegg:
@olivermel One small change, otherwise looks good!  Please make the change as soon as you can, so I can merge the PR and close out the issue, thanks!

tanderegg:
@olivermel One other thing, please open a PR against "develop" instead of "master".  All PR's should be against "develop", I only merge to "master" directly from "develop", as master is used for the Production environment.

PR #106, opened 2016-07-26
ansible intellij role

olivermel:


deploy/deploy_app_server.yml
```diff
@@ -4,7 +4,7 @@
   roles:
     - {role: puppet_disable, when: "disable_puppet"}
     - {role: iptables, when: "iptables_config"}
-    - {role: common, when: "install_glusterfs"}
+#    - {role: common, when: "install_glusterfs"}
```
tanderegg: @olivermel Can you remove the comment here?  It's fine if you disable this while developing, but it should remain enabled in the repository.

deploy/deploy_app_server.yml
```diff
@@ -19,6 +19,7 @@
     - nodejs
     - redis
     - maven
+    - intellij
```
tanderegg: Intellij should be on the Analytics Terminals, not the App Server.

deploy/roles/intellij/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+intellij_version:       2016.1.3
+intellij_url:           "http://download.jetbrains.com/idea/ideaIC-{{intellij_version}}.tar.gz"
+intellij_install_dir:   /opt/idea
+intellij_main_dir:      /usr/local
+inteillij_desktop_dir:  /usr/share/applications
```
tanderegg: Spelling typo here

deploy/roles/intellij/tasks/main.yml
```diff
@@ -0,0 +1,47 @@
+---
+- name: Download intellij
+  get_url: 
+    url: "{{ intellij_url }}"
+    dest: "/tmp/intellij-{{ intellij_version }}"
+  tags: 
+  - intellij
```
tanderegg, line 16: So are we not using the RPM that you built to install IntelliJ?  I think this Ansible role should support that option as well.  There is a variable called "custom_repo" that is True in our 3 environments, and False by default when using Vagrant.  You should have one install command that uses the RPM, and is only run when "custom_repo" is True, and these other commands should only be run when "custom_repo" is False.

tanderegg:
@olivermel A few comments here, plus we should set it up to use the RPM when available.

PR #110, opened 2016-08-03
Neo4j ansible role

olivermel:


deploy/deploy_analytics_terminal.yml
```diff
@@ -21,6 +21,7 @@
     - redis
     - maven
     - intellij
+    - neo4j
```
tanderegg: @olivermel We are not actually installing this on the Analytics Terminal or the App Servers, but rather setting up a new server just for Neo4J.  So you need to create a new entry in the Vagrantfile, and a new deploy_neo4j.yml file for deploying this server.  The playbook should include the usual "common" and "iptables" rules, plus whatever dependencies exist for neo4j (at least the JDK role).

deploy/roles/redis/tasks/main.yml
```diff
@@ -15,6 +15,7 @@
   yum:
```
tanderegg, line 15: This shouldn't be in here, I'm pretty sure.

tanderegg:
Closing in favor of #121

PR #121, opened 2016-08-15
Neo4j ansible role

olivermel:


deploy/deploy_app_server.yml
```diff
@@ -19,6 +19,7 @@
     - nodejs
     - redis
     - maven
+    - neo4j
```
tanderegg: @olivermel Please remove this from the app_server.yml file, it's not needed there.

deploy/deploy_neo4j_server.yml
```diff
@@ -0,0 +1,24 @@
...
+  become: yes 
+  roles:
+    - {role: puppet_disable, when: "disable_puppet"}
+    - {role: iptables, when: "iptables_config"}
+    - {role: common, when: "install_glusterfs"}
+    - {role: epel, when: "use_epel and not 'production' in group_names"}
+    - jdk
+    - neo4j 
+
+- hosts: neo4j_server
```
tanderegg: @olivermel These ports don't need to be open here, we only need the several ports for Neo4J itself, this range of ports was for applications running on app_server instances.

deploy/roles/neo4j/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
+---
+
+- name: Add neo4j user
+  user:
+    name: "{{ neo4j_user }}"
+    system: yes 
+    home: "{{ neo4j_dir }}"
+    shell: "{{ neo4j_user_shell }}"
+    state: present
+    #groups: "{{ neo4j_user_groups | join(',') }} "
```
tanderegg: @olivermel if this isn't needed, please remove the line rather than commenting it out.

deploy/roles/neo4j/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+
+- name: Extract neo4j tarball
+  unarchive:
+    src: "/tmp/neo4j-{{ neo4j_version }}"
+    dest: "{{ neo4j_dir }}"
+    copy: no
+  tags:
+  - neo4j
+  
+- name: Open port for neo4j in iptables
```
tanderegg, line 57: @olivermel we also need the HTTP port, right?

deploy/roles/neo4j/handlers/main.yml
```diff
@@ -0,0 +1,4 @@
+# main.yml
+---
+- name: restart neo4j
+  command: /opt/neo4j/neo4j-community-3.0.4/bin neo4j restart
```
tanderegg: @olivermel Looks like you are missing a slash here, it should be:
`command: /opt/neo4j/neo4j-community-3.0.4/bin/neo4j restart`
That could be the whole problem.  Also, it should actually be:
`command: "{{ neo4j_home_dir }}/bin/neo4j restart`"

deploy/roles/neo4j/defaults/main.yml
```diff
@@ -0,0 +1,42 @@
...
+neo4j_dbms_shell_enabled:                               false
+neo4j_dbms_transaction_timeout:                         60000
+
+neo4j_browser_allow_outgoing_connections:               true 
+neo4j_browser_credential_timeout:                       0m
+neo4j_browser_retain_connection_credentials:            true
+
+
+neo4j_pid_dir:                                          /opt/neo4j/neo4j-community-3.0.4/bin/run
+neo4j_log_dir:                                          /opt/neo4j/neo4j-community-3.0.4/bin/logs
```
tanderegg: You should remove the `/bin` part of both neo4j_pid_dir and neo4j_log_dir

deploy/roles/neo4j/defaults/main.yml
```diff
@@ -0,0 +1,42 @@
...
+neo4j_main_dir:                                         /opt/neo4j/neo4j-community-3.0.4/bin
+neo4j_conf_dir:                                         /opt/neo4j/neo4j-community-3.0.4/conf
+neo4j_nofile_value:                                     40000
+neo4j_update_limits:                                    true
+
+neo4j_dbms_connector_http_type:                         HTTP
+neo4j_dbms_connector_http_enabled:                      true
+neo4j_dbms_connector_http_address:                      localhost:7473
+
+neo4j_bolt_connector:                                   Required
```
tanderegg: I got an error because of this setting, it should be `REQUIRED` not `Required`.  Also, please change the variable name to be `noe4j_bolt_connector_tls_level` so it's clear what this is for.

deploy/roles/neo4j/templates/neo4j-env.sh.j2
```diff
@@ -0,0 +1,2 @@
+export NEO4J_HOME={{ neo4j_main_dir }}
```
tanderegg: This should be set to `neo4j_home_dir`, not `neo4j_main_dir`.  This is why you were getting the ClassPath error.

deploy/roles/neo4j/handlers/main.yml
tanderegg: Come to think of it, we should be running this as an Upstart service or else it could fail at any time.  The upstart config could be as simple as:

```
# Located in /etc/init/neo4j.conf
# Neo4J upstart script
start on runlevel [2345]
stop on runlevel [016]

console output

respawn

exec su -s /bin/sh -c 'exec "$0" "$@"' neo4j -- /opt/neo4j/neo4j-community-3.0.4/bin/neo4j start
```

This should be implemented as a template so the actual path is controlled by {{ neo4j_home_dir }}

tanderegg (changes requested):
@olivermel Great work, just a few issues to address and we should be good to go.

deploy/roles/neo4j/handlers/main.yml
```diff
@@ -1,4 +1,4 @@
 # main.yml
 ---
 - name: restart neo4j
-  command: /opt/neo4j/neo4j-community-3.0.4/bin neo4j restart
+  command: "{{ neo4j_home_dir }}/bin/neo4j restart"
```
tanderegg: @olivermel Now that we have the upstart script, the command should just be `restart neo4j` instead of `{{ neo4j_home_dir }}/bin/neo4j restart`

deploy/roles/neo4j/tasks/main.yml
```diff
@@ -149,6 +149,14 @@
   tags:
   - neo4j
 
+- name: copy neo4j upstart script to /etc/init/neo4j.conf 
+  template: 
+    src: neo4j.upstart.j2 
+    dest: /etc/init/neo4j.conf 
+    mode: 0755
```
tanderegg: @olivermel This should be 644 not 755, since it is a configuration file and not an actual binary.

tanderegg (changes requested):

deploy/roles/neo4j/handlers/main.yml
```diff
@@ -1,4 +1,4 @@
 # main.yml
 ---
 - name: restart neo4j
-  command: "{{ neo4j_home_dir }}/bin/neo4j restart"
+  command: neo4j restart"
```
tanderegg: @olivermel There's an extra trailing `"` at the end now, please remove that.  Can you re-run the script in vagrant by doing:
`vagrant destroy neo4j1`
`vagrant up neo4j1`
To ensure there are no syntax errors?

tanderegg (approved):

PR #139, opened 2016-10-05
Psql update

tanderegg:


frank0631:
:thumbsup: for postgresql95 and sqitch

PR #140, opened 2016-10-07
Stattransfer

olivermel:


deploy/deploy_research_environment.yml
```diff
@@ -4,6 +4,8 @@
   roles:
     - {role: iptables, when: "iptables_config"}
     - {role: common, when: "install_glusterfs"}
+    - stattransfer
```
tanderegg: @olivermel Actually, don't put it here since this would install it on all servers, not just the application servers.  We don't need it on the EOD servers.  Every other group is part of the "research_environment" group, so this is effectively a global group.

PR #152, opened 2016-10-26
Eod

olivermel:


deploy/deploy_research_environment.yml
```diff
@@ -5,6 +5,12 @@
   roles:
     - {role: iptables, when: "iptables_config"}
     - {role: common, when: "install_glusterfs"}
+  tasks:
+    - name: research eod symlinks
+      file:
+         src: []
+         dest: []
+         state: link
```
tanderegg: @olivermel This isn't what I meant, rather use `with_items: eod_symlinks`, and `src: "{{ item.src }}"`, `dest: "{{ item.dest }}`.  Then, in `group_vars/all/main.yml`, add a default setting which is `eod_symlinks: []`.  That way, by default no symlinks are created, but that variable can be overridden like so:

```
eod_symlinks:
  - src: /some/path
    dest: /some/other/path
```

deploy/group_vars/all/main.yml
```diff
@@ -233,3 +233,12 @@ pp_custom_test_scripts:
...
     assertion: "test_output['results'][0]['query_result'] == [[1]]"
+
+# Eod links 
+eod_usr_share_applications:             /usr/share/applications
+eod_usr_share_desktop_directores:       /usr/share/desktop-directories
+eod_etc_gdm_custom_conf:                /etc/gdm/custom.conf 
+
+eod_gnome_menus_applications:           /home/shared/shared_config/gnome_menus/applications
+eod_gnome_menus_desktop_directories:    /home/shared/shared_config/gnome_menus/desktop-directories
+eod_gnome_menus_menus:                  /home/shared/shared_config/gnome_menus/menus
```
tanderegg: @olivermel None of these links should be hardcoded here, they should see my comment above.  These will go in our internal settings repository in the format I described above.

deploy/roles/eod/defaults/main.yaml
```diff
@@ -29,3 +33,6 @@ gconf_settings:
     setting: "/apps/gdm/simple-greeter/disable_user_list"
     type: "bool"
     value: "true"
+
+eod_DisplaysPerHost:        100
+eod_Xconfig_directory:      /usr/local/eod/conf/admin/Xconfig
```
tanderegg: @olivermel Are all of these formatting changes necessary?  Just want to make sure they won't affect the actual output in templates, etc.

deploy/roles/eod/tasks/main.yaml
```diff
@@ -245,3 +252,163 @@
...
+- name: Install yum packages
+  yum:
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - pixmap-miscfixed-font
+    - pixmap-fixed-fonts
+    - abyssinica
+    - tree 
+  when: custom_repo                             ####  ??????
```
tanderegg: @olivermel This when statement isn't needed here, as these packages are not in our custom repository.  They should be available in Vagrant, for instance.

deploy/roles/eod/tasks/main.yaml
```diff
@@ -245,3 +252,163 @@
...
+
+- name: Add Composite to XDMCP_Query.cfg file 
+  lineinfile:                           
+    dest: /usr/local/eod/conf/admin/XDMCP_Query.cfg
+    line: Composite=1                       
+    insertafter: XkbExtension=1                 
+    state: present                          
+  when: cfg_files.stdout == "0"
+  tags:                                     
+    - eod
```
tanderegg: @olivermel What's the significance of cfg_files.stdout == "0"?  Does this mean that if any of the files are found, it will run these scripts on all of them?

deploy/roles/eod/tasks/main.yaml
```diff
@@ -245,3 +252,163 @@
...
+- name: Register symlink3
+  stat: path="{{ eod_gnome_menus_menus }}"
+  register: applic3
+
+- name: Eod link3
+  file:
+    src: "{{ eod_gnome_menus_menus }}"
+    dest: "{{ eod_etc_gdm_custom_conf }}"
+    state: link 
+  when: applic3.stat.exists == True
```
tanderegg: @olivermel See my comment above about how we should be doing this.

deploy/roles/eod/tasks/main.yaml
```diff
@@ -245,3 +252,163 @@
...
+    path: "{{ eod_Xconfig_directory }}"
+    state: directory
+    mode: 0755
+  tags: 
+    - eod 
+
+- name: Set /etc/gdm/gdm.schema file
+  template:
+    src: gdm.schemas.j2
+    dest: /etc/gdm/gdm.schemas
```
tanderegg, line 267: @olivermel It looks like this template doesn't contain any variables.  Are some of these settings supposed to be changeable?  How does this file differ from the default file?

tanderegg (changes requested):

deploy/group_vars/all/main.yml
```diff
@@ -233,3 +233,14 @@ pp_custom_test_scripts:
...
+# Eod symlinks 
+eod_symlinks: []
+
+#eod_usr_share_applications:             /usr/share/applications
+#eod_usr_share_desktop_directores:       /usr/share/desktop-directories
+#eod_etc_gdm_custom_conf:                /etc/gdm/custom.conf 
+#
+#eod_gnome_menus_applications:           /home/shared/shared_config/gnome_menus/applications
+#eod_gnome_menus_desktop_directories:    /home/shared/shared_config/gnome_menus/desktop-directories
+#eod_gnome_menus_menus:                  /home/shared/shared_config/gnome_menus/menus
```
tanderegg: @olivermel Please remove all these lines rather than just commenting them.

tanderegg (changes requested):

deploy/deploy_research_environment.yml
```diff
@@ -4,7 +4,13 @@
   become_method: sudo
   roles:
     - {role: iptables, when: "iptables_config"}
-    - {role: common, when: "install_glusterfs"}
+#    - {role: common, when: "install_glusterfs"}
```
tanderegg: @olivermel Can you remove the comment here?  This line is needed.

deploy/roles/eod/defaults/main.yaml
```diff
@@ -10,14 +10,18 @@ eod_client_ports:
   - 5500
   - 8406
   - 8407
-eod_cm_password: PASSWORD
+  - 6000:6025
+eod_client_ports_udp:
+  - 177
+eod_cm_password:            PASSWORD
```
tanderegg: @olivermel Can you remove the extra white space here?

deploy/roles/eod/tasks/main.yaml
```diff
@@ -245,3 +252,68 @@
...
+    - eod 
+
+- name: Add /usr/local/eod/conf/admin/Xconfig/*.cfg files to register 
+  shell: find /usr/local/eod/conf/admin/Xconfig/ -type f -name \*.cfg -printf "%f\n"
+  register: cfg_files 
+  ignore_errors: yes
+  tags:
+    - eod 
+
+- debug:
```
tanderegg: @olivermel Can you remove the debug lines now that this is working?

deploy/roles/eod/templates/gdm.schemas.j2
```diff
@@ -0,0 +1,130 @@
...
+    <schema>
+      <key>security/DisallowTCP</key>
+      <signature>b</signature>
+      <default>true</default>
+    </schema>
+
+    <schema>
+      <key>greeter/Include</key>
+      <signature>s</signature>
+      <default>{{ eod_entry }}</default>
```
tanderegg: @olivermel Please make the variable name more descriptive, something like eod_greeter_include

tanderegg (changes requested):

tanderegg (approved):

PR #154, opened 2016-10-31
X11 Forwarding on all research app servers

olivermel:


deploy/deploy_research_environment.yml
```diff
@@ -5,6 +5,20 @@
...
+        state: present 
+      when: not custom_repo  
+
+    - name: Ensure X11 Forwarding is present
+      lineinfile:
+        dest: /etc/ssh/sshd_config
+        regexp: "{{ item.regexp }}"
+        line: "{{ item.line }}"
+      with_items:
+        - { regexp: '^X11Forwarding=no', line: 'X11Forwarding=yes' }
```
tanderegg: @olivermel The regex here should be more flexible, since the line may begin with whitespace.  I would use `^\s*X11Forwarding=np` to better capture that.

tanderegg: Also, here you don't need to use with_items, since this will always be a single item.  Just put the values directly in the YAML options.

tanderegg (changes requested):

deploy/deploy_research_environment.yml
```diff
@@ -4,7 +4,28 @@
   become_method: sudo
   roles:
     - {role: iptables, when: "iptables_config"}
-    - {role: common, when: "install_glusterfs"}
+#    - {role: common, when: "install_glusterfs"}
```
tanderegg: @olivermel This line needs to be uncommented.

deploy/deploy_research_environment.yml
```diff
@@ -4,7 +4,28 @@
...
+  tasks:
+    - name: Install libselinux-python and libicu package
```
tanderegg: @olivermel libicu is only needed on the research_matlab server(s), putting it here will install it on all servers in the Research Environment.

tanderegg (changes requested):

PR #157, opened 2016-12-07
Additional packages, symlinks and /etc files for all research env servers

olivermel:


tanderegg:
@olivermel It seems you did not use a new branch for PR #158, so all of these changes are included there.  That's fine, but can you just make sure everything here is in that PR?  Once you've confirmed that, close this one and we'll just use the other one, which I left comments on.

PR #164, opened 2016-12-29
Mesos fixes

mattboehm:
Update Ansible scripts for Mesos machines to fix misc. errors

### DIFF
--- a/deploy/roles/gocd/defaults/main.yaml
+++ b/deploy/roles/gocd/defaults/main.yaml
@@ -1,10 +1,10 @@
 ---
 go_server_address: 10.0.1.26
 go_server_port: 8153
-gocd_agent_package_url: https://download.go.cd/binaries/16.10.0-4131/rpm/go-agent-16.10.0-4131.noarch.rpm
+gocd_agent_package_url: https://download.gocd.io/binaries/16.12.0-4352/rpm/go-agent-16.12.0-4352.noarch.rpm
 gocd_agent_package_name: go-agent-16.10.0-4131.noarch.rpm
 gocd_agent_package_path: /tmp
-gocd_agent_checksum: "a8b811176dad2f5061e5a94a7929a27919b19018"
+gocd_agent_checksum: "1243a5d3066555b08c26e6f90ca31fb567ea7eaf"
 gocd_agent_work_dir: /var/lib/go-agent
 gocd_agent_mem: "128m"
 gocd_agent_max_mem: "256m"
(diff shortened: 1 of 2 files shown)


deploy/roles/gocd/defaults/main.yaml
```diff
@@ -1,10 +1,10 @@
 ---
 go_server_address: 10.0.1.26
 go_server_port: 8153
-gocd_agent_package_url: https://download.go.cd/binaries/16.10.0-4131/rpm/go-agent-16.10.0-4131.noarch.rpm
+gocd_agent_package_url: https://download.gocd.io/binaries/16.12.0-4352/rpm/go-agent-16.12.0-4352.noarch.rpm
 gocd_agent_package_name: go-agent-16.10.0-4131.noarch.rpm
```
tanderegg: @mattboehm Interesting, they changed the URL once again :(  Anyways, the gocd_agent_package_name needs to match the URL, if you wouldn't ming updating that, thanks!

deploy/roles/jdbc/tasks/main.yml
```diff
@@ -1,4 +1,11 @@
 ---
+#Required for computer to make an ssl connection for postgres (since they sign their cert with letsencrypt)
+- name: Install cert utilities
+  yum: 
+    name: ca-certificates
+    state: latest
+  sudo: yes
```
tanderegg: Nice, thanks!  FYI, we've generally been running everything as "sudo" in Aurora, so typically we have not specified sudo per task, but it's not a bad idea.  The newest version of Ansible does this slightly different though, so instead of:
```
sudo: yes
```
it should be:
```
become: yes
become_method: sudo
```

tanderegg (changes requested):

tanderegg:
@mattboehm Ah, sorry I didn't notice: you should work with the "develop" branch primarily, and submit all PR's to "develop".  Basically, "develop" is our working branch, and that is what is run in the Dev environment.  Then, once we're ready to deploy changes to "Test" and "Production", we do a PR from develop -> master.

deploy/roles/jdbc/tasks/main.yml
mattboehm: Thanks, I've updated all the `sudo`'s I've touched in ckan-installer to `become`'s but forgot to correct this one. Since `become_method` defaults to sudo, I've been leaving it off, but perhaps I should be explicitly adding it everywhere I use `become` just to be safe?

mattboehm:
@tanderegg I've made your suggested changes and opened a new PR to merge this into the develop branch instead of master.

PR #173, opened 2017-01-11
Beaker

olivermel:


deploy/deploy_analytics_terminal.yml
```diff
@@ -6,7 +6,7 @@
     - {role: puppet_disable, when: "disable_puppet"}
     - {role: epel, when: "use_epel and not 'production' in group_names"}
     - {role: iptables, when: "iptables_config"}
-    - {role: common, when: "install_glusterfs"}
+#    - {role: common, when: "install_glusterfs"}
```
tanderegg, line 9: @olivermel Please make sure not to include commented out lines in PR's, as if I missed this it could break many things.  If you need to disable the common role, for example, use the EXTRA_VARS environment variable like so:
```
EXTRA_VARS='{ install_glusterfs: "False" }' vagrant up...
```

deploy/roles/beaker/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+beaker_mirror:              "http://d299yghl10frh5.cloudfront.net/beaker-notebook-1.7.1-0-g6dac09a-ubuntu.zip"
+beaker_src_dir:             "/usr/local/src"
+beaker_parent_dir:          "/usr/lib"
```
tanderegg: Let's use /opt instead.  Normally, we've been using /usr/local for applications installed via Yum, and /opt for applications that are extracted directly from an archive.

deploy/roles/beaker/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
...
+beaker_zip_file:            "/usr/local/src/beaker-notebook-1.7.1-0-g6dac09a-ubuntu.zip"
+beaker_main_dir:            "/usr/lib/beaker-notebook-1.7.1-0-g6dac09a-ubuntu"
```
tanderegg: /opt here as well

deploy/roles/beaker/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+- name: Set up beaker environmental variables
+  template:
+    src: beaker-env.sh.j2
+    dest: "/etc/profile.d"
+    mode: 0644
+  tags:
+    - beaker
+
+
+
```
tanderegg: Remove extra newlines at the end of the file here, please.

deploy/roles/beaker/tasks/main.yml
```diff
@@ -0,0 +1,35 @@
...
+  tags:
+    - beaker
+
+- name: Unarchive zipped beaker file
+  unarchive:
+    src: "{{ beaker_zip_file }}"
+    dest: "{{ beaker_parent_dir }}"
+    copy: no
+  tags: 
+    - beaker
```
tanderegg: The "unarchive" command now supports downloading a remote archive and unzipping it all in one command, and I think this would be much cleaner here.  You would add a parameter "remote_src: yes", and remove "copy: no", then set src: to the Beaker URL itself.  This requires Ansible 2.2+, so you might need to update your ansible on the Mac.  If you did this, you could remove the two tasks before this one, and the "beaker_zip_file" and "beaker_src_dir" variables.

deploy/roles/beaker/tasks/main.yml
tanderegg: In the docs (http://beakernotebook.com/getting-started), it mentions an "install-dependencies.sh" script that is designed for Ubuntu.  I doubt it will work out of the box on Centos, so we may have to re-implement that script in Ansible.  Could you take a look at the script and see what's in there, to determine if that is needed?  Also, we may want to create an entry in /usr/share/applications so that Beaker shows up in the Gnome Applications menu.

tanderegg (changes requested):

tanderegg:
Closing, since Beaker has changed into an extension for Jupyter.

PR #181, opened 2017-02-07
Travis

stirlingv:
Added research server to travis ci.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,12 @@ env:
     PLAYBOOK_LOCATION=/aurora/deploy/deploy_mesos_agent.yml 
     EXTRA_VAR_DEF='{"install_glusterfs": false, "install_python": false, "es_message_auth": false, "mesos_master": false, "install_ruby": false, "install_r_updates": false, "install_sqitch": false, "install_r_packages": false, "deploy_iptables": false}'
 
+# gocd  ci_server
+  - >
+    HOST_GROUP=ci_server
+    PLAYBOOK_LOCATION=/aurora/deploy/deploy_gocd.yml 
+    EXTRA_VAR_DEF='{"install_glusterfs": false, "deploy_iptables": false, "install_python": false, "restart_gocd_server": false}'
+
 # gocd ci_agent
   - >
     HOST_GROUP=ci_agent
--- a/.travis.yml
+++ b/.travis.yml
@@ -161,7 +161,7 @@ script:
     docker exec --tty "$(cat ${container_id})" env TERM=xterm sh -c "echo -e \"[${HOST_GROUP}]\\nlocalhost ansible_ssh_host=127.0.0.1 ansible_connection=local\" > /etc/ansible/hosts"
 # Syntax check
   - >
-    docker exec --tty "$(cat ${container_id})" env TERM=xterm ansible-playbook ansible --version --syntax-check "${PLAYBOOK_LOCATION}" --extra-vars "@/tmp/extra_vars.json" 
+    docker exec --tty "$(cat ${container_id})" env TERM=xterm ansible-playbook --syntax-check "${PLAYBOOK_LOCATION}" --extra-vars "@/tmp/extra_vars.json" 
 
 # Test role
   - >
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,12 +49,6 @@ env:
     PLAYBOOK_LOCATION=/aurora/deploy/deploy_mesos_agent.yml 
     EXTRA_VAR_DEF='{"install_glusterfs": false, "install_python": false, "es_message_auth": false, "mesos_master": false, "install_ruby": false, "install_r_updates": false, "install_sqitch": false, "install_r_packages": false, "deploy_iptables": false}'
 
-# gocd  ci_server
-  - >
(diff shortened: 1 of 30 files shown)


.travis.yml
```diff
@@ -67,6 +67,60 @@ env:
...
+  - >
+    HOST_GROUP=research_stata
+    PLAYBOOK_LOCATION=/aurora/deploy/deploy_research_environment.yml 
+    EXTRA_VAR_DEF='{}'
+
+# Research Server research_python 
+  - >
+    HOST_GROUP=research_python
+    PLAYBOOK_LOCATION=/aurora/deploy/deploy_research_environment.yml 
+    EXTRA_VAR_DEF='{"install_python": false, "clean_postgresql": false, "install_r_packages": false, "install_r_updates": false}'
```
tanderegg: clean_postgresql should be false by default, was that not the case here?

deploy/deploy_gocd.yml
```diff
@@ -5,7 +5,7 @@
   roles:
     - {role: common, when: "install_glusterfs"}
     - {role: puppet_disable, when: "disable_puppet"}
-    - {role: iptables, when: "iptables_config"}
```
tanderegg: Oops, it occurred to me that we do in fact need this variable, as it controls whether or not to run the iptables commands in Dev/Test/Prod, where we usually do not want to.  Perhaps keep this as it was, and use a new variable inside the role instead of using the same variable?  Then leave iptables_config as True, and set that new one to False in Travis.

deploy/roles/gocd/meta/main.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 dependencies:
   - {role: python, when: "install_python and custom_repo"}
-  - {role: python-build, when: "install_python and not custom_repo"}
+  - {role: python-build, when: "not custom_repo"}
```
tanderegg, line 4: This shouldn't be changed, right?

tanderegg (changes requested):
Added a few comments, also it looks like Travis did not run, we'll have to look into why that is.

deploy/roles/gocd/meta/main.yaml
stirlingv: @tanderegg We only need to skip a few python tasks to avoid the 10 minute limitation in travis. So rather than skipping the whole role, I thought it made sense to only skip the tasks that take a bit to install.

deploy/deploy_gocd.yml
stirlingv: :thumbsup:

.travis.yml
stirlingv: I think I needed to add that for the postgres build. I will verify

stirlingv:
@tanderegg I think this PR is ready to be merged. It includes research server, data warehouse, and the fixed pentaho role.

.travis.yml
```diff
@@ -94,7 +161,7 @@ script:
     docker exec --tty "$(cat ${container_id})" env TERM=xterm sh -c "echo -e \"[${HOST_GROUP}]\\nlocalhost ansible_ssh_host=127.0.0.1 ansible_connection=local\" > /etc/ansible/hosts"
 # Syntax check
   - >
-    docker exec --tty "$(cat ${container_id})" env TERM=xterm ansible-playbook --syntax-check "${PLAYBOOK_LOCATION}" --extra-vars "@/tmp/extra_vars.json" 
+    docker exec --tty "$(cat ${container_id})" env TERM=xterm ansible-playbook ansible --version --syntax-check "${PLAYBOOK_LOCATION}" --extra-vars "@/tmp/extra_vars.json"
```
tanderegg: I don't think the "ansible --version" part should be in there, right?

deploy/deploy_gocd.yml
tanderegg: We still need the `when: iptables_config` wherever the `iptables` role is used.

tanderegg (changes requested):

.travis.yml
stirlingv: Ha, you're right. I just put that in there when I was trying to determine what commands we were able to use. Is there any harm in leaving it, just to add clarity, or would you prefer I remove it? @tanderegg

deploy/deploy_gocd.yml
stirlingv: Found it... Taken care of. Checking in code in moments

.travis.yml
tanderegg: Please remove it, I'm not sure why it doesn't error as written, I would think that it would!

deploy/roles/gocd/meta/main.yaml
tanderegg: Cool

deploy/deploy_research_environment.yml
```diff
@@ -3,7 +3,7 @@
   become: yes
   become_method: sudo
   roles:
-    - {role: iptables, when: "iptables_config"}
+    - iptables
```
tanderegg: @stirlingv We need to keep the conditional here as well

tanderegg (changes requested):

PR #184, opened 2017-02-22
Gnome

olivermel:


deploy/deploy_research_environment.yml
```diff
@@ -111,21 +111,30 @@
...
+      with_items: "{{ python3_packages }}"
+    
+    - name: Install Gnome-Terminal packages 
+      yum:
+        name: "{{ item }}"
+        state: present 
+        disable_gpg_check: yes 
+      with_items:
+        - "@X Window System"
+        - "@Desktop"
```
tanderegg: These packages are already installed via the "eod" group.  I think the problem that we're trying to address is that gnome-terminal is not installed on some of the Application servers, and when a program like Stata tries to open it, it is not found.  So we don't need the whole @X Window and @Desktop packages, but rather just gnome-terminal.

tanderegg (changes requested):

PR #185, opened 2017-02-22
Kibana fix

stirlingv:
This branch is not named well. These fixes update when conditions in a number of roles to only run with deploy_iptables is true, for integration with Travis-CI


tanderegg:
@stirlingv What's the difference between this PR and the "Travis" one?  it seems to be the same changes, except that the "Traivs" PR was rebased?

PR #186, opened 2017-02-27
Research env installs

rioh:


tanderegg:
@rioh Looks like this one now needs a rebase, thanks!

PR #201, opened 2017-03-29
Logs proxy server

stirlingv:


stirlingv:
Looks like the log permission are being set correctly. Example:
```
[vagrant@proxy log]$ getfacl krb5libs.log 
# file: krb5libs.log
# owner: root
# group: root
user::rw-
group::r--
group:admin:r-x
mask::r-x
other::---
```

tanderegg:
@stirlingv Looks like the Travis failure is a real one, I think you need to install the "acl" package when in Travis, since the Docker base image appears not to have it.

deploy/roles/apache/tasks/main.yml
```diff
@@ -1,12 +1,9 @@
...
-  get_url:
-    url: "{{ apache_24_repo }}"
-    dest: /etc/yum.repos.d
-    mode: 644
-    validate_certs: False
-  when: not custom_repo
+- name: Add Apache repo
+  yum: 
+    name: centos-release-scl
+    state: present
```
tanderegg: @stirlingv You still need the `when: not custom_repo` conditional

deploy/roles/apache/tasks/main.yml
```diff
@@ -51,3 +48,7 @@
 - name: Save rules
   command: "/sbin/service iptables save"
   when: iptables_config
+
+- name: create certs for apache localhost
+  command: openssl req -x509 -nodes -days 365 -newkey rsa:2048 -subj "/C=US/ST=CFPB/L=Washington/O=Dis/CN=www.cf.gov" -keyout /etc/pki/tls/private/localhost.key -out /etc/pki/tls/certs/localhost.crt
+  when: not custom_repo
```
tanderegg: I would add `creates: /etc/pki/tls/certs/localhost.crt`, this ensures it does not recreate the cert if it already exists.

deploy/roles/common/tasks/main.yml
```diff
@@ -73,3 +73,8 @@
   yum:
     name: screen
     state: present
+
+- name: Create admin admin group
```
tanderegg: admin twice here

deploy/roles/apache/tasks/main.yml
tanderegg: Also, we need a section to setup the permissions logs for Apache as well.  By default they are /var/log/httpd/error_log and /var/log/httpd/access_log, but since we are installing an alternate version they may be in /var/log/httpd24/

tanderegg (changes requested):

deploy/roles/apache/tasks/main.yml
tanderegg: It might make sense to set default ACL's on the whole folder, since in theory other logs could end up there.  If set the default via ACL's, any new files will have those permissions.

stirlingv:
:+1:

tanderegg:
@stirlingv Last thing needed, per this comment: Also, we need a section to setup the permissions logs for Apache as well. By default they are /var/log/httpd/error_log and /var/log/httpd/access_log, but since we are installing an alternate version they may be in /var/log/httpd24/

tanderegg:
@stirlingv Looks like you need to update your develop branch with the upstream's, then rebase one more time, thanks!

tanderegg (approved):

deploy/roles/apache/tasks/main.yml
```diff
@@ -51,3 +49,20 @@
...
+- name: Allow admins to read logs
+  acl:
+    path: "{{ item }}"
+    entity: "{{ restricted_group_name }}"
+    etype: group
+    permissions: rx
+    state: present
+  with_items:
+    - /var/log/httpd24/
+    - /var/log/httpd/
```
tanderegg: @stirlingv This should actually be controlled by a variable - Rather than doing both httpd and httpd24, this depends on which one is installed.  If we are using httpd24 (which is the current default), then /var/log/httpd won't exist.  So rather than doing "with_items" for both, just replace `{{ item }}`  with a new variable `{{ apache_log_folder }}` with a default of  `/var/log/httpd24`.  Also, we should have `recursive: yes` and `default: yes` to ensure that all files currently in that folder are readable by the admin user, and all new files ever created will also be readable.

deploy/roles/r/defaults/main.yml
```diff
@@ -22,3 +24,9 @@ r_env_vars:
   CC: /opt/rh/devtoolset-2/root/usr/bin/gcc
   CPP: /opt/rh/devtoolset-2/root/usr/bin/cpp
   CXX: /opt/rh/devtoolset-2/root/usr/bin/c++
+
+r_packages_with_opts:
+  -
+    name: udunits2
```
tanderegg: Minor quibble, but the `name:` key can go on the same line as the `-`, which improves readability a bit.

tanderegg (changes requested):
@stirlingv Sorry, a few last changes, then we're done!

deploy/roles/apache/tasks/main.yml
tanderegg: Also, could you try `rX` for the permissions?  In the normal setfacl command, using a capital X instead of lowercase x will only make directories executable, and not files.  It would be nice if this works in Ansible too, but it's not the end of the world if log files end up being executable.

tanderegg (changes requested):

stirlingv:
@tanderegg changes included, and rebased commits. Let me know!

tanderegg:
@stirlingv Looks like it needs to be rebased once more, sorry about that!

PR #204, opened 2017-04-07
Added devtools6 role

rioh:


deploy/roles/devtools6/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+    name: "{{ item }}"
+    state: present
+  with_items:
+    - devtoolset-6-binutils
+    - devtoolset-6-build
+    - devtoolset-6-gcc
+    - devtoolset-6-gcc-c++
+    - devtoolset-6-gcc-gfortran
+    - devtoolset-6-gdb
+    # - devtoolset-6-git
```
tanderegg: Is there no git package in devtools 6?

deploy/roles/devtools6/templates/devtools-6.sh.j2
```diff
@@ -0,0 +1 @@
+source /opt/rh/devtoolset-6/enable
```
tanderegg, line 1: Doesn't look like this template is actually used anywhere.  It might make sense to place it in /opt to make it easy to get to for users, and ensure the file mode is 755.

tanderegg (changes requested):

deploy/roles/devtools6/tasks/main.yml
rioh: Doesn't appear to be, unless I'm not looking in the right places. I couldn't find any information regarding the removal.

tanderegg: Sounds good, you can just remove the commented out line then.

deploy/roles/devtools6/templates/devtools-6.sh.j2
rioh: Ok, added!

deploy/roles/devtools6/tasks/main.yml
rioh: done!

deploy/roles/devtools6/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+  when: not custom_repo
+  tags:
+    - devtools
+
+- name: Copy devtools shell script
+  template:
+    src: devtools-6.sh.j2
+    dest: /opt/devtools-6.sh
+    mode: 0755
+  when: not custom_repo
```
tanderegg: @rioh We don't need the conditional here, since the script should be copied in both Vagrant and the actual environment.

tanderegg (changes requested):
One last minor change, thanks!

rioh:
Removed the conditional. thanks

PR #208, opened 2017-04-25
Added python34 setup tools and easy_install pip to python34 role

rioh:


deploy/roles/python3/tasks/main.yml
```diff
@@ -1,7 +1,17 @@
-# main.yml - Install Python 2.7, either from a repo or a built RPM
+# main.yml - Install Python 3.4, either from a repo or a built RPM
 ---
 - name: Install Python34 from repo
   yum:
     name: python34
     state: present
     disable_gpg_check: yes
```
tanderegg: @rioh You can actually remove this disable_gpg_check as well, since we are now installing Python34 from EPEL and not our own repo.

tanderegg (changes requested):

deploy/roles/python3/tasks/main.yml
rioh: @tanderegg Done!

tanderegg (approved):

PR #212, opened 2017-05-02
Fix Travis Timeout, and Allow Admins to read logs in Mesos Cluster

stirlingv:


deploy/roles/elasticsearch/tasks/main.yml
```diff
@@ -45,15 +45,15 @@
 
 - name: Open ports for Elasticsearch transport communication
   command: "iptables -I {{ iptables_chain }} 3 -m state --state NEW -p tcp --dport {{ es_transport_ports }} -j ACCEPT"
-  when: iptables_config
+  when: iptables_config and proxy_server_deployed
```
tanderegg: @stirlingv We don't need to add the variable here, since this command doesn't rely on the proxy server.

tanderegg (changes requested):

deploy/roles/elasticsearch/tasks/main.yml
```diff
@@ -45,15 +45,15 @@
...
 
 - name: Open ports for Elasticsearch http communication
   command: "iptables -I {{ iptables_chain }} 3 -m state --state NEW -p tcp --dport {{ es_http_ports }} -s {{ hostvars[groups['proxy_server'][0]]['ansible_ssh_host'] }} -j ACCEPT"
-  when: iptables_config
+  when: iptables_config and proxy_server_deployed
```
tanderegg: @stirlingv Rather than creating a new variable, could you just say `iptables_config and 'proxy_server' in groups`?

deploy/roles/elasticsearch/tasks/main.yml
```diff
@@ -45,15 +45,15 @@
...
 
 - name: Save rules
   command: "/sbin/service iptables save"
-  when: iptables_config
+  when: iptables_config and proxy_server_deployed
```
tanderegg: @stirlingv You don't need the variable here either.

deploy/roles/mesos_master/tasks/main.yml
```diff
@@ -548,3 +548,32 @@
...
     - mesos-master
+
+- name: Create empty log files if they don't exist
+  file:
+    path: "{{ item }}"
+    state: touch
+    owner: root
+    group: root
+    mode: 0640
+  with_items: "{{ log_file_locations }}"
```
tanderegg: @stirlingv These folders should always exist, and if they don't we do not want files here instead, so this section can be removed.

deploy/roles/elasticsearch/defaults/main.yml
```diff
@@ -1,6 +1,7 @@
 ---
-es_version: 2.3
+es_version: 2.x
 es_full_version: 2.3.4
 es_data_directory: /opt/elasticsearch/data
 es_transport_ports: "9300:9400"
 es_http_ports: "9200:9299"
+proxy_server_deployed: true
```
tanderegg: @stirlingv See below, this variable can probably be removed, but if we did want to keep it it should live in group_vars/all/main.yml, since it is a "global" variable not just related to Elasticsearch but rather the state of the Enclave as a whole.

deploy/roles/elasticsearch-config/tasks/main.yml
```diff
@@ -153,3 +153,13 @@
...
+
+- name: Allow admins to read logs
+  acl:
+    path: "{{ item }}"
+    entity: "{{ restricted_group_name }}"
+    etype: group
+    permissions: rx
+    state: present
+  with_items:
+    - /var/log/elasticsearch
```
tanderegg: @stirlingv This should have `recursive: yes` to ensure that any existing log files are readable, and it should have a separate section for setting the default, as with the other roles.

deploy/roles/common/tasks/main.yml
```diff
@@ -80,3 +80,27 @@
...
     state: present
+    
+- name: Create empty log files if they don't exist
+  file:
+    path: "{{ item }}"
+    state: touch
+    owner: root
+    group: root
+    mode: 0640
+  with_items: "{{ log_file_locations }}"
```
tanderegg, line 91: @stirlingv /var/log/messages will always exist on Centos 6, so you don't need this block here.

deploy/deploy_research_environment.yml
```diff
@@ -107,15 +111,12 @@
     - postgresql-client
     - r
     - r-studio-desktop
-    - r-studio-server-config
     - r-studio-server
+    - r-studio-server-config
```
tanderegg, line 115: @stirlingv Good call, this was obviously backwards for some reason...

deploy/roles/common/tasks/main.yml
tanderegg: @stirlingv Bump on this one if you didn't see it, thanks!

deploy/roles/kerberos/tasks/main.yml
```diff
@@ -51,3 +51,4 @@
     recursive: yes
     state: present
   with_items: "{{ log_file_locations }}"
```
tanderegg: @stirlingv I think we might want to "namespace" these variables in general, so we don't run into the same issue we saw with elasticsearch.  So this one could be `kerberos_log_file_locations`.

deploy/roles/log-courier/tasks/main.yml
```diff
@@ -50,8 +50,7 @@
     owner: root
     group: root
     mode: 0640
-  with_items:
-    - "{{ log_file_locations }}"
+  with_items: "{{ log_file_locations }}"
```
tanderegg: `logcourier_log_file_locations`

deploy/roles/mesos_master/defaults/main.yml
```diff
@@ -4,6 +4,10 @@
 # Number of log files to keep at a time
 zookeeper_maxbackupindex: 20
 
+log_file_locations:
```
tanderegg: `mesos_master_log_file_locations`

deploy/roles/mesos_master/tasks/main.yml
```diff
@@ -548,3 +548,23 @@
...
+
+- name: Set recursive admin permissions on log folders
+  acl:
+    path: "{{ item }}"
+    entity: "{{ restricted_group_name }}"
+    etype: group
+    permissions: rX
+    recursive: yes
+    state: present
+  with_items: "{{ log_file_locations }}"
```
tanderegg: `mesos_master_log_file_locations`

deploy/roles/mesos_master/tasks/main.yml
```diff
@@ -548,3 +548,23 @@
...
+
+- name: Set default admin permissions on log folders
+  acl:
+    path: "{{ item }}"
+    entity: "{{ restricted_group_name }}"
+    etype: group
+    permissions: rX
+    default: yes
+    state: present
+  with_items: "{{ log_file_locations }}"
```
tanderegg: ditto

deploy/roles/mesos_master/defaults/main.yml
tanderegg: Also, we need to add the permissions tasks to the mesos_agent role as well, although there it is just `/var/log/mesos` and not `/var/log/zookeeper`.  Call that one `mesos_agent_log_file_locations`

tanderegg (changes requested):
Sorry @stirlingv, a few last changes!

deploy/roles/common/tasks/main.yml
stirlingv: Sorry, I missed that. Standby for update @tanderegg

deploy/roles/kerberos/tasks/main.yml
stirlingv: @tanderegg makes sense - will do and update PR

deploy/roles/common/tasks/main.yml
stirlingv: Hey @tanderegg are you sure this will always exist? See travis failure - https://travis-ci.org/stirlingv/aurora/jobs/229248009 

Is this stood up as part of file_server_1 or 2? - Does this mean we should skip this in travis?

stirlingv:
@tanderegg is the messages folder created as part of file_server_1 or 2? When I remove creating the empty folder, it says it doesn't exist. https://travis-ci.org/stirlingv/aurora/jobs/229248009 Should we skip this task in travis?

tanderegg:
@stirlingv It's supposed to be created by the OS itself on all servers, perhaps the Docker image does not include it though... I supposes it is safe to include code that ensures that it exists, just make sure the ownership is root:root, mode 640.

tanderegg:
@stirlingv Also, it should not be a folder but rather a file.  It is basically the OS's primary log file.

deploy/roles/mesos_master/tasks/main.yml
```diff
@@ -548,3 +548,23 @@
...
+
+- name: Set default admin permissions on log folders
+  acl:
+    path: "{{ item }}"
+    entity: "{{ restricted_group_name }}"
+    etype: group
+    permissions: rX
+    default: yes
+    state: present
+  with_items: "{{ mesos_master_log_file_locations }}"
```
tanderegg, line 570: @stirlingv We also need to add these permissions tasks to the mesos_agent role, but only for /var/log/mesos and not /var/log/zookeeper.  Thanks!

deploy/roles/elasticsearch-config/tasks/main.yml
```diff
@@ -153,3 +153,23 @@
...
+
+- name: Set default admin permissions on log folders
+  acl:
+    path: "{{ item }}"
+    entity: "{{ restricted_group_name }}"
+    etype: group
+    permissions: rX
+    default: yes
+    state: present
+  with_items: "{{ es_log_file_locations }}"
```
tanderegg, line 175: @stirlingv Since you're already doing Elasticsearch, let's do the Logstash and Kibana as well, which have their own roles, and logs at /var/log/logstash and /var/log/kibana respectively.  Thanks!

tanderegg (changes requested):

deploy/roles/elasticsearch-config/tasks/main.yml
stirlingv: @tanderegg Will do! Standby

deploy/roles/mesos_master/tasks/main.yml
stirlingv: 👍

stirlingv:
@tanderegg hold off on merging this for a second. Looks like there is a regression issue, and the location of the zookeeper logs isn't right. Investigating now

.travis.yml
```diff
@@ -35,19 +35,19 @@ env:
   - >
     HOST_GROUP=elasticsearch
     PLAYBOOK_LOCATION=/aurora/deploy/deploy_elastic_search.yml 
-    EXTRA_VAR_DEF='{"skip_glusterfs": true, "install_python": false, "es_message_auth": false}'
+    EXTRA_VAR_DEF='{"skip_glusterfs": true, "install_python": false, "proxy_server_deployed": false}'
```
tanderegg: @stirlingv I think "proxy_server_deployed" is no longer needed, right?

.travis.yml
```diff
@@ -35,19 +35,19 @@ env:
...
 
 # elk 
   - >
     HOST_GROUP=elk
     PLAYBOOK_LOCATION=/aurora/deploy/deploy_elk.yml 
-    EXTRA_VAR_DEF='{"skip_glusterfs": true, "install_python": false, "es_message_auth": false, "deploy_iptables": false}'
+    EXTRA_VAR_DEF='{"skip_glusterfs": true, "install_python": false, "deploy_iptables": false, "proxy_server_deployed": false}'
```
tanderegg: @stirlingv same here

tanderegg (changes requested):

tanderegg (approved):

PR #214, opened 2017-05-10
Added sas files

rioh:


.travis.yml
```diff
@@ -107,7 +107,7 @@ env:
   - >
     HOST_GROUP=research_sas
     PLAYBOOK_LOCATION=/aurora/deploy/deploy_research_environment.yml 
-    EXTRA_VAR_DEF='{"install_python": false, "restart_gocd_server": false}'
+    EXTRA_VAR_DEF='{"install_python": false, "restart_gocd_server": false, "install_sas": false}'
```
tanderegg: @rioh Because `install_sas` is defaulted to `false` in the role, it technically isn't needed to be explicitly set here.

deploy/roles/sas/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+- name: Install yum dependencies
+  yum:
+    name: libXext.i686
+    state: present
+  tags:
+    - sas
+
+#- name: Stop SAS Servers
+
+#- name: Backup files
```
tanderegg: @rioh Are these place holders for future work, or are they no longer needed?

deploy/roles/sas/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+  lineinfile:
+    dest: "{{ sas_config_home }}/appserver/studio/bin/setenv.sh"
+    regexp: 'JVM_OPTS='
+    line: 'JVM_OPTS="{{ java_heap_size }}"'
+    backup: yes
+  when: install_sas
+  tags:
+    - sas
+
+#- name: Restore backup
```
tanderegg: @rioh Same here, if not needed we can remove.

deploy/roles/sas/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+
+- name: Copy over pam.d config from login (usually works)
+  shell: cp -f /etc/pam.d/login /etc/pam.d/sasauth
+  tags:
+    - sas
+
+- name: Add rta init.d daemon script
```
tanderegg: Should be `add SAS init.d`... right?

tanderegg (changes requested):
@rioh Just a few minor issues, thanks!

deploy/roles/sas/tasks/main.yml
rioh: They're placeholders, but it's easy enough to add back in, so I'll take it out.

PR #215, opened 2017-05-11
Allow admins to read EOD logs

stirlingv:
@tanderegg This will fail the travis build, since it doesn't include all the regression changes included in the travis_timeout branch. It will also likely have merge conflicts in the travis.yml file and the deploy_research_environment.yml (the changes in this are to remove what looks like a typo) files. I should be able to resolve those easily once the travis_timeout branch is merged in.


deploy/roles/eod/tasks/main.yaml
```diff
@@ -49,10 +49,27 @@
   tags:
     - eod
 
+- name: Create admin group
+  group:
+    name: "{{ restricted_group_name }}"
+    state: present
+
```
tanderegg: @stirlingv Isn't this group already created in the `common` role?

deploy/roles/eod/tasks/main.yaml
```diff
@@ -49,10 +49,27 @@
...
+    mode: 0640
+
+- name: Set recursive admin permissions on log file
+  acl:
+    path: "{{ eod_install_directory }}/log"
+    entity: "{{ restricted_group_name }}"
+    etype: group
+    permissions: rX
+    default: yes
+    state: present
```
tanderegg, line 76: @stirlingv Don't we need to sections here, one with `default: yes` and the other `recursive: yes`?

tanderegg (changes requested):

deploy/roles/eod/tasks/main.yaml
stirlingv: @tanderegg It is, but in travis we don't call the deploy_research host before running the research_eod host. So the common role is not called. 
```
- hosts: research_eod
  become: yes
  become_method: sudo
  roles:
    - eod
    - odbc
```

Maybe we should add a when condition to only add the group when running in travis?

tanderegg: @stirlingv The common role is applied in the `hosts: research_environment` block at the top of the playbook, which applies to all servers in the RE, so you can safely remove this block, thanks!

deploy/roles/eod/tasks/main.yaml
tanderegg: @stirlingv I meant "two sections" not "to sections" :)

stirlingv: @tanderegg I am looking at this now - When I first did this there was a reason I didn't have a recursive parameter. I'll keep ya posted...

tanderegg (approved):

PR #220, opened 2017-05-30
Added tasks for admin access to logs

rioh:


tanderegg:
@rioh Looks like the Travis build failed on the SAS playbook because the "admin" group doesn't exist, probably because we need to make sure the Ansible inventory file includes the "[research_environment]" gropu, which triggers the "common" role.

rioh:
yup! I was just waiting for stirlingv's changes to get merged in to develop

PR #221, opened 2017-06-02
Allow Admin group to read gocd logs

stirlingv:


deploy/roles/gocd/defaults/main.yaml
```diff
@@ -17,3 +17,6 @@ gocd_rpms:
   - vim
   - emacs
   - "@Development tools"
+
+gocd_log_file_locations:
+  - /var/log
```
tanderegg: @stirlingv We shouldn't do all logs in this folder, just the ones specific to our applications.  The GoCD agent's logs (`gocd` role) go in /var/log/go-agent, and the GoCD server's logs (role `gocd_server`) go in /var/log/go-server.

tanderegg (changes requested):

tanderegg (approved):

PR #225, opened 2017-06-14
Final adjustments for sas ansible scripts

rioh:
Added install_sas when and notify clauses to some sas tasks
Changed from using shell to copy module for pam
Added meta folder to add shared tasks

### DIFF
--- a/deploy/roles/sas/tasks/main.yml
+++ b/deploy/roles/sas/tasks/main.yml
@@ -3,6 +3,9 @@
   yum:
     name: libXext.i686
     state: present
+  when: install_sas
+  notify:
+    - restart sas-server
   tags:
     - sas
 
@@ -18,6 +21,7 @@
   template:
     src: sas_template.txt.j2
     dest: "{{ sas_response_file_path }}"
+  when: install_sas
   tags:
     - sas
 
@@ -27,6 +31,8 @@
   register: installer_cmd
   failed_when: installer_cmd.stderr != ""
   when: install_sas
+  notify:
+    - restart sas-server
   tags:
     - sas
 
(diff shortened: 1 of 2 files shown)


deploy/roles/sas/tasks/main.yml
```diff
@@ -37,11 +43,16 @@
     line: 'JVM_OPTS="{{ java_heap_size }}"'
     backup: yes
   when: install_sas
```
tanderegg: @rioh I think we should remove the `when: install_sas` conditional here, since it may be the case that we want to change the JVM heap size, but not run the full install script.

deploy/roles/sas/tasks/main.yml
```diff
@@ -50,6 +61,7 @@
     src: sas_init.d_template.j2
     dest: /etc/init.d/sas-studio
     mode: 0751
+  when: install_sas
```
tanderegg: @rioh Here too we probably don't want the conditional, as we may want to modify the init script.  One thing that might be worth adding, for example, is a `status` command that is common on init scripts to see if the program is running or not.  This should have a `notify` block though, to restart the process if the script is changed.

deploy/roles/sas/tasks/main.yml
```diff
@@ -85,3 +97,9 @@
   with_items: "{{ sas_log_file_locations }}"
   tags:
     - sas
+
+- name: Enable sas service
+  service:
+    state: started
+    enabled: yes
+  when: install_sas
```
tanderegg: @rioh You don't need the conditional here, since if the service is already started and enabled, this will be a no-op.  It doesn't hurt to ensure these things are true on every run.

tanderegg (changes requested):

PR #246, opened 2017-06-30
Adds python34-develop and postgres_client tags

rioh:


stirlingv (approved):
Looks good to me :+1:

PR #262, opened 2017-08-11
Added members_of directive to restrict by group

rioh:


deploy/roles/r-shiny-config/templates/shiny-server.conf.j2
```diff
@@ -31,6 +31,9 @@ server {
     run_as :HOME_USER:;
     user_dirs;
 
+    # Restrict to only allow certain groups to host peronal applications.
+    members_of {{ r_shiny_groups }};
```
tanderegg: @rioh Have you tested this blank?  I'm just worried it might prevent anyone from hosting an app if there is no value there.  It might make sense to add a conditional around this, saying `if r_shiny_groups != ""`?

deploy/roles/r-shiny-config/defaults/main.yml
```diff
@@ -5,3 +5,5 @@ shiny_required_users: "admin"
 shiny_auth_proxy: False
 shiny_auth_proxy_header: ""
 shiny_admin_port: 4151
+
+r_shiny_groups: ""
```
tanderegg: @rioh - The rest of the variables are just "shiny_" instead of "r_shiny_", I don't have a strong preference either way, but they should probably be consistent.

deploy/roles/r-shiny-config/templates/shiny-server.conf.j2
rioh: Yep, having a blank allows all users to be able to host

tanderegg:
@rioh FYI the reason Travis failed is related to the Postgres 9.6 scripts, and totally unrelated from this PR, so I merged, thanks!

PR #267, opened 2017-09-25
Stattransfer

stirlingv:


tanderegg:
@stirlingv This too could use a rebase, thanks!

deploy/roles/stattransfer/templates/license1282.txt
```diff
@@ -1,8 +1,8 @@
 [Registration]
 PRODUCT=Stat/Transfer
-NAME=Research & Markets Team
-ORGANIZATION=CFPB
-MAJOR-VERSION=13
+NAME={{stattransfer_name}}
+ORGANIZATION={{stattransfer_organization}}
+MAJOR-VERSION={{stattransfer_major_version}}
```
tanderegg, line 5: @stirlingv Where is the `stattransfer_major_version` variable set?  I don't see it in defaults/main.yml above.

PR #279, opened 2017-10-23
Added task to remove installed files

rioh:


deploy/roles/stata/tasks/main.yaml
```diff
@@ -8,6 +8,18 @@
...
+  file:
+    path: "{{ stata_install_directory }}/{{ item }}"
+    state: absent
+  with_items:
+    - installed.150
+    - stata
+    - xstata
+    - isstata.150
+  tags:
+    - stata
```
tanderegg, line 21: @rioh What about the license file, stata.lic?  If that is there, the installer prompts will change which will interfere with the expect script, no?  Or does removing these 4 files prevent that check from even happening?

rioh: Yes, I think the installed.150 prevents it from running that check. We can remove the license too, it won't hurt.

PR #291, opened 2017-10-30
Updated epel path

rioh:


tanderegg:
@rioh Should I let the Travis run on this one to verify it works as expected?

PR #293, opened 2017-11-16
Added task to add shiny-server conf

rioh:


deploy/roles/r-shiny/templates/shiny-server.conf.j2
```diff
@@ -0,0 +1,29 @@
...
+    i=0
+    while [ $i -lt 5 ]
+    do
+        pgrep "shiny-server" || exit 1
+        sleep 10
+        i=$((i+1))
+    done
+end script
+
+exec shiny-server --pidfile=/var/run/shiny-server.pid >> /var/log/shiny-server.log 2>&1
```
tanderegg: @rioh I'm not sure that it is strictly necessary, but it might be a good idea to wrap this line in `script` and `end script` to ensure it runs after pre-start but before post-start.  This may work anyways, but being explicit could help with some edge cases or weirdness down the line.

tanderegg (changes requested):

deploy/roles/r-shiny/templates/shiny-server.conf.j2
```diff
@@ -0,0 +1,31 @@
...
+
+start on runlevel [2345]
+stop on runlevel [016]
+
+limit nofile 1000000 1000000
+
+post-stop exec sleep 3
+
+pre-start script
+  exec sleep 100
```
tanderegg, line 13: @rioh Should we have a `end script` here too?

tanderegg (changes requested):

PR #296, opened 2017-11-27
Opensource Postgres and Citus

rioh:


.travis.yml
```diff
@@ -151,7 +151,7 @@ env:
     PARENT_GROUP=all
     HOST_GROUP=db_server
     PLAYBOOK_LOCATION=/aurora/deploy/deploy_db_server.yml
-    EXTRA_VAR_DEF='{"deploy_iptables": false, "install_ppas": false, "skip_glusterfs": true}'
+    EXTRA_VAR_DEF='{"deploy_iptables": false, "install_postgres": false, "skip_glusterfs": true}'
```
tanderegg: @rioh Ideally, we should be able to remove this flag since we were only setting it to False to avoid trying to install the EnterpriseDB version of Postgres.  Now that we can deploy the open-source Postgres, we should be able to re-enable this in Travis.

deploy/group_vars/all/main.yml
```diff
@@ -29,7 +29,7 @@ shared_folders:
 home_directory: "/testorg"
 
 # Location for importing data into PPAS
-pp_file_transfer_dir: "/home/work/dw/file_transfers"
+pp_file_transfer_dir: "/home/dtwork/dw/file_transfers"
```
tanderegg: This should actually be "/home/work", since that is the folder setup in Vagrant.

deploy/group_vars/all/main.yml
```diff
@@ -226,8 +229,6 @@ pp_custom_test_scripts:
     script: "SELECT 1;"
     assertion: "test_output['results'][0]['query_result'] == [[1]]"
 
-db_automation_custom_packages: []
```
tanderegg, line 229: This is a default variable that is only used internally, so it should be kept.

deploy/roles/citus/tasks/main.yml
```diff
@@ -0,0 +1,48 @@
...
+  become: yes
+  become_method: sudo
+  lineinfile:
+    dest: "{{ pp_datadir }}/postgresql.conf"
+    regexp: "^shared_preload_libraries*"
+    line: "shared_preload_libraries = '{{ pp_shared_preload_libraries }}'"
+  notify:
+    - restart postgres
+  tags:
+    - citus
```
tanderegg: This variable is already set in the postgresql.conf template file in the postgresql-server role.  I think it makes sense to explicitly set pp_shared_preload_libraries='citus' in the Vagrantfile, rather than edit the postgresql.conf file again here.

deploy/roles/citus_coordinator/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+  postgresql_exec:
+    login_host: localhost
+    login_user: "{{ pp_superaccount }}"
+    login_password: "{{ pp_superpassword }}"
+    port: "{{ pp_serverport }}"
+    db: postgres
+    script: "SELECT * from master_add_node('{{ item }}', {{ pp_serverport }})"
+  check_mode: no
+  with_items: "{{ citus_worker_nodes }}"
+  when: install_postgres
```
tanderegg: I think we can remove the "when: install_postgres" part from here.

deploy/roles/citus_coordinator/tasks/main.yml
```diff
@@ -0,0 +1,25 @@
...
+  become_user: "{{ pp_superaccount }}"
+  postgresql_exec:
+    login_host: localhost
+    login_user: "{{ pp_superaccount }}"
+    login_password: "{{ pp_superpassword }}"
+    port: "{{ pp_serverport }}"
+    db: postgres
+    script: "SELECT * from master_get_active_worker_nodes()"
+  check_mode: no
+  when: install_postgres
```
tanderegg: Same here.

tanderegg: Also, it might be worth it to register the result of this query, and use a Ansible "assert" block to verify the output, see here: http://docs.ansible.com/ansible/latest/assert_module.html

deploy/roles/jdk/tasks/main.yml
```diff
@@ -1,4 +1,9 @@
 ---
+- name: Load variables
+  include_vars:
+    file: rhel_7.yml
+  when: "{{ ansible_distribution_major_version }} == 7"
```
tanderegg: If we do it this way, the below paths will be hardcoded and cannot be overridden by our internal configs.  I think ultimately we will just need to make sure `jre_lib_path` and `java_home_path` are set correctly manually.  To get things working properly in Vagrant, for now it might be easiest to override these in the Vagrantfile itself via extra-vars.

deploy/roles/postgresql-server-config/tasks/main.yml
```diff
@@ -3,67 +3,110 @@
...
+  become_method: sudo
+  lineinfile:
+    dest: "{{ pp_datadir }}/pg_hba.conf"
+    regexp: '^(.*)::1/128(.*)trust'
+    backrefs: yes
+    line: '\1::1/128\2md5'
+  when: install_postgres
+  notify: restart postgres
+
+- meta: flush_handlers
```
tanderegg: This didn't seem to be necessary before, did something go wrong with the postgres user when deploying?

deploy/roles/postgresql-server/tasks/main.yml
```diff
@@ -1,10 +1,46 @@
...
+    name: 'openssl'
+    state: latest
+  tags:
+    - postgresql-server
+
+- name: Install Postgres pgdg
+  yum:
+    name: "{{ pgdg_repository }}"
     state: present
-  when: not 'production' in group_names
```
tanderegg: I think we should keep the `where:` clause, but change it to `when: not custom_repo`, which is the general flag to indicate whether the Redhat Satellite server is available (i.e. in our internal environment).

deploy/roles/postgresql-server/templates/postgresql.conf.96.j2
```diff
@@ -0,0 +1,637 @@
...
+#---------------------------------------------------------------------------
+# RUNTIME INSTRUMENTATION AND TRACING
+#---------------------------------------------------------------------------
+
+#timed_statistics = {{ pp_timed_statistics }}			# record wait timings, defaults to on
+
+# Add settings for extensions here
+{% for setting in pp_extension_settings %}
+{{ setting.name }} = {{ setting.value }}
+{% endfor %}
```
tanderegg: Don't we need to have sections for the edb variables somewhere in this file?  Wrapped with `{{ if install_edb }}`?

deploy/roles/postgresql-server/templates/pg_hba.conf.j2
```diff
@@ -1,6 +1,6 @@
 {{ template_masthead }}
 # TYPE  DATABASE        USER            ADDRESS                 METHOD
 
-{% for setting in pp_hba_settings %}
+{% for setting in pg_hba_settings %}
```
tanderegg, line 4: Since we've changed some variables from pp_XXX to pg_XXX, we should probably change all of them, and prefix the edb ones with edb_

.travis.yml
tanderegg: Expanding on this, we may want to just remove the `install_postgres` variable completely from here (since the default is to install open-source Postgres, it should now work on Travis), and add tags to the various tasks instead.

deploy/Vagrantfile
```diff
@@ -583,6 +582,104 @@ Vagrant.configure(2) do |config|
...
+        ansible.playbook = "deploy_citus_server.yml"
+        ansible.inventory_path = "vagrant_hosts"
+        ansible.tags = ansible_tags
+        ansible.verbose = ansible_verbosity
+        ansible.extra_vars = ansible_extra_vars
+        ansible.limit = ansible_limit
+      end
+    end
+  end
+
```
tanderegg: Can you add a db_server_edb box with the proper overrides to install EnterpriseDB instead of OSS?

deploy/Vagrantfile
```diff
@@ -583,6 +582,104 @@ Vagrant.configure(2) do |config|
...
+        ansible_extra_vars = ENV['DB_SERVER_EXTRA_VARS']
+      else
+        print "To use the db_server VM, set the DB_SERVER_EXTRA_VARS environment variable to the name of a local YAML file that contains pp_web_username (email addresses only) and pp_web_password.\n"
+        print "Example:\n"
+        print "$ export DB_SERVER_EXTRA_VARS=local_settings.yml\n"
+        print "$ cat local_settings.yml\n"
+        print "---\n"
+        print "pp_web_username: myusername@example.com\n"
+        print "pp_web_password: mypassword\n"
+      end
```
tanderegg: This section is unnecessary when installing the open source version, since `pp_web_username` and `pp_web_password` are only needed for EDB.

deploy/group_vars/all/main.yml
```diff
@@ -210,6 +210,9 @@ gocd_service_password: password
 ### PostgreSQL Settings ###
 #=========================#
 
+# Whether or not to do anything at all
+install_postgres: True
```
tanderegg: See above about removing this now that it should work on Travis.

deploy/roles/epel/defaults/main.yml
```diff
@@ -1,2 +1,2 @@
 epel_path: http://dl.fedoraproject.org/pub/epel
-epel_rpm: epel-release-latest-6.noarch.rpm
+epel_rpm: epel-release-latest-{{ ansible_distribution_major_version }}.noarch.rpm
```
tanderegg, line 2: :+1:

tanderegg (changes requested):
Thanks for all this Risa!  Just a few things to sort out, and I'll try actually running in Vagrant tomorrow.

deploy/roles/postgresql-server-config/tasks/main.yml
rioh: Before, I had changed the pg_hba.conf entry to 'trust' to enable postgres to be able to login since postgres user doesn't have a password initially. This way I can change it back to md5 and the rest of the playbook will still work, since we added a password.

deploy/group_vars/all/main.yml
rioh: This variable was defined twice in this file so I just removed this one

deploy/roles/citus/tasks/main.yml
rioh: But then we would have to remember to add it as an extra-var when we deploy to prod?

deploy/group_vars/all/main.yml
tanderegg: Ah, that makes sense, thanks!  Probably a rebasing issue.

deploy/roles/citus/tasks/main.yml
tanderegg: We should be able to just add it to aurora_internal once, for the "citus" group_vars, that way we can use additional extensions with Citus too.

deploy/roles/postgresql-server-config/tasks/main.yml
tanderegg: Ah, got it, thanks!

deploy/roles/citus/tasks/main.yml
rioh: ok, got it

deploy/Vagrantfile
```diff
@@ -583,6 +584,128 @@ Vagrant.configure(2) do |config|
...
+    cc_config.vm.provision "shell", inline: <<-SHELL
+      # bug with vagrant 1.9+ and centos/7
+      sudo ifup eth1
+    SHELL
+
+    cc_config.vm.provision "ansible" do |ansible|
+      ansible_extra_vars = {
+        "jre_lib_path": "/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.131-11.b12.el7.x86_64/jre/lib",
+        "java_home_path": "/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.131-11.b12.el7.x86_64/",
+        "pp_shared_preload_libraries": "citus"}
```
tanderegg: @rioh The way the postgresql-server role is setup currently, the "pp_install" variable must be set to True the first time a PostgreSQL instance is setup.  This is possible using `EXTRA_VARS="{pp_install: True}" vagrant up ...`, but since you are overriding the `ansible_extra_vars` variable, it doesn't work.  Could you update this (and the others), to instead merge the variables with the existing `ansible_extra_vars` variable?  In Ruby, it's like this:
```
ansible_extra_vars.merge({ jre_lib_path: "...", ... })
```
Note the slight change in notation, the key is not quoted in this case, which is the slightly more "Rubyist" way of doing things (that's not a real word, Ruby doesn't have an equivalent to "Pythonic" it seams :)).

deploy/deploy_citus_server.yml
```diff
@@ -0,0 +1,58 @@
...
+        regexp: '[[0-9+]\.]+\s+{{ item }}'
+        line: "{{ hostvars[item]['ansible_ssh_host'] }} {{ item }}"
+      with_items: "{{ groups['citus'] }}"
+      when: not custom_repo
+
+
+- hosts: citus
+  roles:
+    - role: postgresql-server-config
+    - role: citus
```
tanderegg: I had to do some rearranging here.  Basically, since the "citus" library is added as part of the postgresql.conf template now, Postgres won't start unless Citus is already installed.  So, I had to remove the task for starting the postgresql server from the postgresql-server role, and put it in the playbook instead (which would have to be done in all playbooks using the role).  Then, I had to move the citus role up to run right after postgresql-server, and then postgresql-server-config after that, all in one `roles` list.  I then merged the `tasks` and `post_tasks` into a single `post_tasks` block.  This ensures that the handler isn't run until after Citus has been installed, since that would fail as well.  Let me know if this is unclear!

deploy/roles/citus/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+
+- name: Add pgpass file
+  become: yes
+  become_user: "{{ pp_superaccount }}"
+  template:
+    src: pgpass.j2
+    dest: ~/.pgpass
+    mode: "0600"
+  tags:
+    - citus
```
tanderegg, line 38: One thing I forgot to mention from the Design Document, we are planning to use CGroups to isolate Citus, which will allow for future multipurpose use of these servers.  This should be relatively straightforward thanks to SystemD on RHEL 7, here's the relevant docs: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/resource_management_guide/sec-Modifying_Control_Groups#sec-Modifying_Unit_Files.  I'm thinking we could have two variables, `citus_mem_limit` and `citus_cpu_limit`, which would be undefined by default, and the relevant tasks would have a clause such as: `when: citus_mem_limit is defined and citus_cpu_limit is defined`.  We could then test with `EXTRA_VARS='{citus_mem_limit: 1024, citus_cpu_limit: 50}'` or whatever is appropriate when using Vagrant, and define them in Production in group_vars.

tanderegg (changes requested):
A few issues I uncovered while testing in Vagrant, although I'm still not done since I couldn't get the coordinators and workers to talk to each other.  Investigating that issue, could be a local config problem, but wanted to give you initial feedback in the mean time.

deploy/Vagrantfile
rioh: That makes sense. I've just been updating the default value in the roles to test it.

deploy/Vagrantfile
```diff
@@ -629,9 +631,7 @@ Vagrant.configure(2) do |config|
     SHELL
 
     server.vm.provision "ansible" do |ansible|
-      ansible_extra_vars = {
-        "jre_lib_path": "/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.131-11.b12.el7.x86_64/jre/lib",
-        "java_home_path": "/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.131-11.b12.el7.x86_64/"}
+      ansible_extra_vars = ansible_extra_vars.merge({ jre_lib_path: "/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.131-11.b12.el7.x86_64/jre/lib", java_home_path: "/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.131-11.b12.el7.x86_64/"})
```
tanderegg: @rioh I know this is nitpicky, but would you please put each variable on a separate line like they were before (i.e. newline after the first `{` and newline before the last `}`.  This makes it much easier to see and review, without having to deal with line wrapping.  Same goes for the lines below, thanks!

deploy/deploy_citus_server.yml
```diff
@@ -10,6 +10,8 @@
     - {role: log-courier, when: "install_logstash"}
     - role: jdk
     - role: postgresql-server
+    - role: citus
+    - role: postgresql-server-config
   tasks:
```
tanderegg: Could you make this `post_tasks` instead?  I'm not sure that it really matters, but it may help to be explicit to ensure no future regressions in the execution order.

deploy/roles/citus/tasks/main.yml
```diff
@@ -34,3 +34,26 @@
     mode: "0600"
   tags:
     - citus
+
+- name: Ensure systemd directory exists
+  file:
+    state: directory
+    path: /etc/systemd/system/postgresql-9.6.service.d
```
tanderegg, line 43: Was this path not present when Postgres 9.6 was installed?  It should have already been there, just want to verify.

tanderegg: Although I could be wrong, I'm still getting familiar with the inner workings of SystemD, it's possible this directory is only created if needed.

deploy/roles/citus/tasks/main.yml
```diff
@@ -34,3 +34,26 @@
...
+  tags:
+    - citus
+
+- name: Add CGroup settings cpu limit
+  template:
+    dest: /etc/systemd/system/postgresql-9.6.service.d/CPUShares.conf
+    src: citus_cpu.conf.j2
+  when: citus_cpu_limit is defined
+  tags:
+    - citus
```
tanderegg: @rioh Could you add `notify` blocks to these sections, to ensure that PostgreSQL is restarted to make these take effect?

deploy/upgrade_db_server.yml
```diff
@@ -34,6 +34,14 @@
   roles:
     - role: postgresql-server
   post_tasks:
+    - name: Ensure PG is started
+      service:
+        name: "{{ pp_servicename }}"
+        state: started
+        enabled: yes
+      tags:
+        - postgresql-server
```
tanderegg: @rioh This is actually already done on line 51 below, it is just missing the `enabled: yes` part, could you add that in and remove this block?

tanderegg (changes requested):
@rioh Some last (hopefully) issues, thanks Risa!

deploy/Vagrantfile
rioh: yes sorry, hasty commit

deploy/roles/citus/tasks/main.yml
rioh: I thought so too, but it wasn't there and my job failed. There's probably a better way to force it to create it.

deploy/roles/citus/tasks/main.yml
```diff
@@ -0,0 +1,61 @@
+---
+- name: Add Citus rpm
+  get_url:
+    url: "{{ rpm_url }}"
+    dest: /tmp/rpm.sh
+    mode: "0755"
+  tags:
```
tanderegg: @rioh Relatively minor, but this should have `when: not custom_repo` as well.

deploy/roles/citus/tasks/main.yml
```diff
@@ -0,0 +1,61 @@
...
+    chdir: /tmp/
+  when: not custom_repo
+  tags:
+    - citus
+
+- name: Install Citus
+  become: yes
+  become_method: sudo
+  yum:
+    name: "{{ citus_pkg }}"
```
tanderegg: @rioh Since the Enterprise version has some additional packages, can you change `citus_pkg` to `citus_pkgs`, which is a list instead using `with_items`?  That way we can easily adjust.

tanderegg (changes requested):

.travis.yml
```diff
@@ -151,7 +151,7 @@ env:
     PARENT_GROUP=all
     HOST_GROUP=db_server
     PLAYBOOK_LOCATION=/aurora/deploy/deploy_db_server.yml
-    EXTRA_VAR_DEF='{"deploy_iptables": false, "install_ppas": false, "skip_glusterfs": true}'
+    EXTRA_VAR_DEF='{"deploy_iptables": false, "skip_glusterfs": true}'
```
tanderegg: @rioh Can you add `pp_install: true`?  This is required in order to cause the role to run the `initdb` task, without that Travis will fail.  Hopefully this will finally allow Travis to run through completely to verify our Postgres scripts!

tanderegg:
@rioh Looks like Travis got much farther, but failed on an execution of my postgresql_exec module.  Looks like it failed to play nice with the "register" option in Ansible, likely due to some change in the most recent Ansible version.  I'll go ahead and merge, we can tackle that later!

PR #297, opened 2017-11-29
updating UltraEdit to version 16.22

stirlingv:


deploy/roles/ultra-edit/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+ue_repo_name_version_16_18: UltraEdit-16.1-0.18.x86_64
+ue_repo_name_version_3_3: UltraEdit-3.3-0.12.x86_64
+ue_repo_name_version_16_22: UltraEdit-16.1-0.22.x86_64
+ue_repo_name_version_15_8: UltraEdit-15.1-0.8.x86_64
```
tanderegg: @stirlingv It might make sense to have this be a single list variable instead of 4 separate variables, called `ue_uninstall_list` or something, then just do `with_items: {{ ue_uninstall_list }}` below.

deploy/roles/ultra-edit/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+  with_items:
+    - "{{ ue_repo_name_version_16_18 }}"
+    - "{{ ue_repo_name_version_3_3 }}"
+    - "{{ ue_repo_name_version_16_22 }}"
+    - "{{ ue_repo_name_version_15_8}} "
+  when: custom_repo
+
+- name: Install Ultra Edit
+  yum:
+    name: "{{ ue_repo_name_version_16_22 }}" #UltraEdit.x86_64
```
tanderegg: Regarding above comment, this variable would become `ue_install_version`, or something along those lines.

tanderegg (changes requested):

deploy/roles/ultra-edit/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
+- name: Uninstall Ultra Edit
+  yum: 
+    name: "{{ item }}" #UltraEdit.x86_64
+    state: removed
+    disable_gpg_check: yes
+  with_items:
+    - "{{ ue_uninstall_list }}"
```
tanderegg: I think this should just be `with_items: "{{ ue_uninstall_list }}"`, although it may work either way.  This way is probably a bit cleaner.

tanderegg (changes requested):
@stirlingv One minor change, then I will merge, thanks!

deploy/roles/ultra-edit/tasks/main.yml
stirlingv: @tanderegg refactored, rebased, and pushed to ultraedit branch

PR #299, opened 2017-12-07
Added variables for r core and packages state

rioh:


deploy/roles/r/tasks/main.yml
```diff
@@ -9,7 +9,7 @@
 - name: Install R-core
   yum:
     name: R-core
-    state: latest
+    state: "{{ r_core_state }}"
```
tanderegg, line 12: I think it probably makes sense just to have one variable, r_core_state, which controls both of these tasks.  I can't think of a scenario where we would want these two settings to be different, since R-devel depends on R-core anyways, so we should just rely on a single variable.  Does that makes sense to you?

rioh: Sure!

PR #301, opened 2017-12-11
Added pgdg key task

rioh:


deploy/roles/postgresql-server/tasks/main.yml
```diff
@@ -31,6 +31,13 @@
   tags:
     - postgresql-server
 
+- name: Install Postgres key
+  rpm_key:
+    key: "{{ postgres_repository_key_url }}"
+    state: present
+  tags:
+    - postgres_server
```
tanderegg: @rioh Should be `postgresql_server`

PR #302, opened 2017-12-12
R monolithic to roles

lel99999:
Initial breakdown into roles.


deploy/roles/r-core/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+  file:
+    dest: /usr/lib64/R/etc/{{ item }}
+    mode: 0644
+  with_items:
+    - ldpaths
+    - Makeconf
+  tags:
+    - r-core
+
+- include_role:
```
tanderegg: @lel99999 Actually, I think we should keep this role separate, as there may be times that we want to run the `r-core` tasks without also running the `r-libs` tasks.

deploy/roles/r-libs/defaults/main.yml
```diff
@@ -0,0 +1,34 @@
...
+  - QuantLib-devel
+  - libcurl-devel
+  - libxml2-devel
+  - hunspell-devel
+  - udunits2
+  - udunits2-devel
+
+r_packages:
+  - "devtools"
+#  - "http://cran.r-project.org/src/contrib/Archive/RQuantLib/RQuantLib_0.3.12.tar.gz" # The latest version requires QuantLib 1.4.0, which is not supported by Centos 6.5
```
tanderegg: @lel99999 Let's also remove this line and the QuantLib dependencies above, since it has issues and we can use "devtools" for Vagrant testing anyways.

deploy/roles/r-core/meta/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+dependencies:
+  - {role: epel, when: "use_epel and not 'production' in group_names"}
+  - devtools
+  - shared
+  - common
```
tanderegg: `common` can be removed here, since it is not a strict dependency of `r-core`, just `r-libs`

deploy/roles/r-update/defaults/main.yml
```diff
@@ -0,0 +1,34 @@
...
+  - QuantLib-devel
+  - libcurl-devel
+  - libxml2-devel
+  - hunspell-devel
+  - udunits2
+  - udunits2-devel
+
+r_packages:
+  - "devtools"
+#  - "http://cran.r-project.org/src/contrib/Archive/RQuantLib/RQuantLib_0.3.12.tar.gz" # The latest version requires QuantLib 1.4.0, which is not supported by Centos 6.5
```
tanderegg: Same here, we can just remove the Quantlib stuff.

deploy/roles/r-update/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+    - r-update
+
+#Address - No /usr/bin/Rscript error
+#- name: Install Rscript
+#  become: yes
+#  become_method: sudo
+#  yum:
+#    name: https://mirrors.sonic.net/epel/6/x86_64/epel-release-6.8.noarch.rpm
+#    state: present
+#  register: rscript_epel
```
tanderegg: If this isn't needed, it should be removed rather than just commented out (and we should install EPEL via the `epel` role, so this shouldn't be needed)

deploy/roles/r-update/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+- name: Install packages required by R packages
+  become: yes
+  become_method: sudo
+  yum:
+    name: "{{ item }}"
+    state: present
+    disable_gpg_check: yes
+  with_items: "{{ r_dependencies }}"
+  tags:
+    - r-update
```
tanderegg: This section should probably be in `r-libs`, since there may be times that that role is run without updating first, but new rpm dependencies were introduced.  It might be easier to remove the separate role for `r-update`, and includ the update tasks in `r-libs`, and go back to using the `when: update_r_packages` conditional, or you could just move the `include_role` task to after this task is executed.

deploy/roles/r-update/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+#  become_method: sudo
+#  yum:
+#    name: https://mirrors.sonic.net/epel/6/x86_64/epel-release-6.8.noarch.rpm
+#    state: present
+#  register: rscript_epel
+
+- include_role:
+    name: epel
+  tags:
+    - epel
```
tanderegg: @lel99999 This section shouldn't be needed, since the "epel" role is included as a dependency of this role and so should execute first.

deploy/roles/r-update/tasks/main.yml
```diff
@@ -0,0 +1,67 @@
...
+
+- name: Install R
+  become: yes
+  become_method: sudo
+  yum:
+    name: R
+    state: present
```
tanderegg: This should be removed, since R is installed in the `r-core` role.  It can be assumed that `r-core` has been run at least once before `r-update` is ever run, since we don't want to include it as a dependency explicitly, which would cause it to run whenever `r-update` is run.

tanderegg (changes requested):
Thanks @lel99999, looks good overall, just have a few organizational changes to make.  In addition to those, could you move the Citus stuff into a separate PR?  When you do, could you also have those scripts live in the `test` folder rather than the `deploy` folder?  Thanks!

lel99999:
Have made the requested changes.  I've kept the r-update role as a placeholder role for future considerations.

deploy/deploy_r_withroles_tester.yml
```diff
@@ -0,0 +1,7 @@
+---
+- hosts: analytics_terminal
+  name: install r with separated roles
+  roles:
+#   - r-core
+#   - r-libs
+    - r-update
```
tanderegg: @lel99999 If the testing is done, this should be removed.

deploy/roles/r-core/tasks/main.yml
```diff
@@ -0,0 +1,197 @@
...
+  file:
+    dest: /usr/lib64/R/etc/{{ item }}
+    mode: 0644
+  with_items:
+    - ldpaths
+    - Makeconf
+  tags:
+    - r-core
+
+#- include_role:
```
tanderegg: Everything below this line should be removed, since it is now in the r_libs role, right?

deploy/roles/r-libs/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+    - rlibs
+
+#Address - No /usr/bin/Rscript error
+#- name: Install Rscript
+#  become: yes
+#  become_method: sudo
+#  yum:
+#    name: https://mirrors.sonic.net/epel/6/x86_64/epel-release-6.8.noarch.rpm
+#    state: present
+#  register: rscript_epel
```
tanderegg: Commented out sections should just be removed.

deploy/roles/r-update/tasks/main.yml
```diff
@@ -0,0 +1,46 @@
...
+  become: yes
+  become_method: sudo
+  command: /usr/bin/Rscript --slave --no-save --no-restore-history -e "update.packages(repos=c('http://cran.rstudio.com/'),ask=FALSE,checkBuilt=TRUE);"
+  environment: "{{ r_env_vars }}"
+  async: 2400
+  poll: 60
+  tags:
+    - r-update
+
+ when: install_r_updates
```
tanderegg: This needs to be at the same indent level as the parameters above it, or it will result in syntax error.

tanderegg (changes requested):
Hi Luan - Thanks for your work on this, a few cleanup items needed still, plus removing the Citus stuff, then I can merge.

deploy/roles/r-core/tasks/main.yml
tanderegg: @lel99999 Bump, can you make this change?

deploy/roles/r-libs/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+  become_method: sudo
+  file:
+    path: "{{ item.path }}"
+    state: absent
+  with_items:
+    - "{{ locked_dirs.files }}"
+  tags:
+    - r-update
+
+- include_role:
```
tanderegg: @lel99999 We don't want to include the role in here, that is already covered by the playbooks and dependencies.

deploy/roles/r-libs/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+- name: Update existing R packages
+  become: yes
+  become_method: sudo
+  command: /usr/bin/Rscript --slave --no-save --no-restore-history -e "update.packages(repos=c('http://cran.rstudio.com/'),ask=FALSE,checkBuilt=TRUE);"
+  environment: "{{ r_env_vars }}"
+  async: 2400
+  poll: 60
+  tags:
+    - r-update
+
```
tanderegg: There should not be a newline here.

deploy/roles/r-libs/tasks/main.yml
```diff
@@ -0,0 +1,111 @@
...
+- name: Install R packages
+  command: /usr/bin/Rscript --slave --no-save --no-restore-history -e "if (! ('{{ item }}' %in% installed.packages()[,'Package'])) { install.packages(pkgs='{{ item }}', INSTALL_opts=c('--no-test-load'),repos=c('http://cran.rstudio.com/'),dep=TRUE); print('Added'); } else { print('Already installed'); }"
+  with_items: "{{ r_packages }}"
+  register: r_result
+  failed_when: "r_result.rc != 0 or 'had non-zero exit status' in r_result.stderr"
+  changed_when: "'Added' in r_result.stdout"
+  environment: "{{ r_env_vars }}"
+  tags:
+    - r-libs
+# when: install_r_packages
```
tanderegg: This line can be removed, along with the other commented out lines below.

deploy/roles/r-update/tasks/main.yml
tanderegg: @lel99999 It seems like you made this change in r-libs, but did not remove r-update, can you remove the r-update files since they are no longer needed?

deploy/roles/r-core/defaults/main.yml
```diff
@@ -0,0 +1,34 @@
...
+  CXX: /opt/rh/devtoolset-2/root/usr/bin/c++
+
+r_packages_with_opts:
+  - name: udunits2
+    include: /usr/include/udunits2
+    lib: /usr/lib64
+
+r_system_package_dir:
+  - /usr/lib64/R/library
+  - /usr/share/doc
```
tanderegg: @lel99999 I missed this before, but actually all of these variables can be removed from `r-core/defaults`, since they are not used in this role any more.  See below the `roles/r-core/tasks/main.yml` file, none of them are used.

tanderegg (changes requested):

deploy/roles/r-core/tasks/main.yml
tanderegg: @lel99999 Bump again, please remove the `include_role` part, thanks!

deploy/roles/r-libs/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+- name: Install packages required by R packages
+  become: yes
+  become_method: sudo
+  yum:
+    name: "{{ item }}"
+    state: present
+    disable_gpg_check: yes
+  with_items: "{{ r_dependencies }}"
+  tags:
+    - r-update
```
tanderegg: @lel99999 The `tag:` normally should just match the role name to keep it simple, so `r-libs` instead of `r-update`.

deploy/roles/r-libs/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+  when: install_r_updates
```
tanderegg: @lel99999 Remove `when: install_r_updates`, as that is no longer needed.  We can just run `--exclue r-libs` now to exclude the whole thing by tag.

deploy/roles/r-libs/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+  become: yes
+  become_method: sudo
+  find:
+    paths: "{{ r_system_package_dir }}"
+    pattern: '00LOCK-*'
+    file_type: directory
+  register: locked_dirs
+  tags:
+    - r-update
+  when: install_r_updates
```
tanderegg: @lel99999 Remove the `when` conditional here as well.

deploy/roles/r-libs/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+- name: Update existing R packages
+  become: yes
+  become_method: sudo
+  command: /usr/bin/Rscript --slave --no-save --no-restore-history -e "update.packages(repos=c('http://cran.rstudio.com/'),ask=FALSE,checkBuilt=TRUE);"
+  environment: "{{ r_env_vars }}"
+  async: 2400
+  poll: 60
+  tags:
+    - r-update
+  when: install_r_updates
```
tanderegg: @lel99999 Remove the `when` here as well.

deploy/roles/r-libs/defaults/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+install_r_updates: true
```
tanderegg: @lel99999 See below, this is no longer needed.

deploy/roles/r-libs/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+
+#--------------------------------------------------------------- RUpdate ^ - R-Libs v
```
tanderegg: @lel99999 We can remove this logical separation, as we always want to update packages before installing new ones.  Let's just consider everything to be `r-libs`.

tanderegg (commented):
@lel99999 Some additional changes, please.  Also, there should be a `roles/r-core/defaults/main.yml` file, which basically just has the `r_core_state` variable in it, and the original `r` task should be updated just to have an `include_role: r-core, r-libs` for now.

deploy/roles/r-core/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+r_core_state: ""
```
tanderegg: @lel99999 This should be `r_core_state: present` by default, to prevent R from updating unintentionally.

deploy/roles/r-core/defaults/main.yml
```diff
@@ -0,0 +1,8 @@
...
+
+- include_role: 
+    name: "{{ item }}"
+  with_items:
+    - r-core
+    - r-libs
```
tanderegg: @lel99999 Sorry, I didn't mean here, `defaults/main.yml` is just for default variables.  `r-core` and `r-libs` should be listed in the playbooks themselves, outside of the roles, like in https://github.com/cfpb/aurora/blob/develop/deploy/deploy_r_terminal.yml where it current has just `- r`.

deploy/roles/r-core/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+  lineinfile:
+    dest: "~/.bashrc"
+    line: umask 0022
+    regexp: (.*)umask(.*)
+    state: present
+
+- name: Install R-core
+  yum:
+    name: R-core
+    state: latest
```
tanderegg: This should be `state: "{{ r_core_state }}"` to allow us to control when R is updated.

deploy/roles/r-core/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
...
+  when: "{{ (rstudio_installed is defined and rstudio_installed.results|length > 0 and item['yumstate']=='installed')}}"
+  with_items: "{{ rstudio_installed.results | default([]) }}"
+  notify: restart rstudio-server
+  tags:
+    - r-core
+
+- name: Install base packages
+  yum:
+    name: "{{ item }}"
+    state: latest
```
tanderegg: @lel99999 This should also use `state: "{{ r_core_state }}"` to make sure `R-devel` doesn't update `R-core` unintentionally.

deploy/roles/r-core/meta/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+dependencies:
+  - {role: epel, when: "use_epel and not 'production' in group_names"}
+  - devtools
```
tanderegg: @lel99999 `devtools` should be removed, as it is not a depdency of `r-core`, just `r-libs`

deploy/roles/r-core/tasks/main.yml
```diff
@@ -0,0 +1,65 @@
+---
+- name: Ensure umask is correct
+  lineinfile:
+    dest: "~/.bashrc"
+    line: umask 0022
+    regexp: (.*)umask(.*)
+    state: present
```
tanderegg, line 7: @lel99999 This should have the `tags: r-core` parameter, or else it might not get run.

tanderegg (changes requested):
@lel99999 A few last things, almost there :)

deploy/deploy_app_server.yml
```diff
@@ -78,6 +78,8 @@
   roles:
     - jdk
     - r
+    - r-core
+    - r-libs
```
tanderegg: @lel99999 One last thing to do here, remove the `- r` role from these playbooks, since `r-core` and `r-libs` replaces it.

tanderegg (changes requested):

deploy/deploy_app_server.yml
tanderegg: @lel99999 I see that the "r" role was removed below, but it is still in this playbook.  Can you check all playbooks and make sure the `- r` is replaced by `- r-core` and `- r-libs` in each one?  Thanks!

deploy/deploy_research_environment.yml
```diff
@@ -118,6 +118,8 @@
     - {role: python-libs}
     - postgresql-client
     - r
```
tanderegg: @lel99999 `- r` role still present here.

tanderegg (changes requested):

deploy/deploy_app_server.yml
lel99999: grepped app_server, gocd, mesos_agent, research_env ...; should be good to go now

tanderegg (approved):

PR #303, opened 2017-12-18
Citus fixes

tanderegg:


tanderegg:
@rioh Can you review?  Thanks!

deploy/roles/postgresql-server-config/tasks/main.yml
```diff
@@ -16,13 +16,16 @@
   become_method: sudo
   lineinfile:
     dest: "{{ pp_datadir }}/pg_hba.conf"
-    regexp: '^(.*)::1/128(.*)md5'
+    regexp: '^host(.*)postgres(.*)md5'
     backrefs: yes
-    line: '\1::1/128\2trust'
+    line: 'host\1postgres\2trust'
```
rioh, line 21: Will these work locally?

deploy/roles/shared/handlers/main.yml
```diff
@@ -31,13 +31,17 @@
...
-- name: restart postgres
-  become: yes
-  service:
-    name: "{{ pp_servicename }}"
-    state: restarted
+#- name: restart postgres
+#  become: yes
+#  service:
+#    name: "{{ pp_servicename }}"
+#    state: restarted
```
rioh, line 38: Would this work if we were to deploy postgres without postgres-config (--tags=postgres) or that wouldn't happen?

deploy/roles/postgresql-server-config/tasks/main.yml
tanderegg: Added a line to the defaults for pg_hba.conf to make this work in Vagrant.

deploy/roles/shared/handlers/main.yml
tanderegg: Handlers should run even though they don't have the tag, as long as a task that does have the tag is marked "changed"

PR #304, opened 2017-12-19
update to latest version of r-studio

stirlingv:
This is where I found the latest open source RPM: https://www.rstudio.com/products/rstudio/download-server/


deploy/roles/r-studio-server/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
-rstudio_server_package_name: http://download2.rstudio.org/rstudio-server-rhel-1.0.143-x86_64.rpm
+rstudio_server_package_name: https://download2.rstudio.org/rstudio-server-rhel-pro-1.1.383-x86_64.rpm
```
tanderegg: @stirlingv In Aurora, we should use the open-source version as the default, otherwise we won't be able to run it out of the box with "vagrant up".  We can override to the commercial version in Prod.  For now, test the non-pro version.  Thanks!

PR #308, opened 2018-01-03
Citus minor bug fixes

rioh:


deploy/roles/shared/handlers/main.yml
```diff
@@ -40,6 +40,7 @@
 # rhel7
 - name: reload-daemon
   systemd:
+    name: "{{ pp_servicename }}"
```
tanderegg, line 43: @rioh I think I added this as a generic "reload-daemon" command that would work for multiple roles.  I'm not sure that adding the `name` parameter here will do anything, since there is no `state` parameter to indicate what to do with the named service.  Or, perhaps I'm misunderstanding how reload-daemon works, and a specific service must be specified?

rioh: @tanderegg Oh I see.  Well according to the docs, that should have worked then, but it complained when I tried it locally.  Let me see what's happening.

PR #309, opened 2018-01-03
Citus updates

rioh:


deploy/roles/postgresql-server-config/tasks/main.yml
```diff
@@ -240,3 +240,32 @@
     type: group
     state: present
     admin_option: no
+
+- name: Set permissions on postgres conf file
+  acl:
+    path: "{{ pg_conf_file }}"
+    entity: "{{ item }}"
+    etype: group
+    permissions: rX
```
tanderegg: This can just be "r", since this file should never be executable.

deploy/roles/shared/handlers/main.yml
```diff
@@ -40,6 +40,7 @@
 # rhel7
 - name: reload-daemon
   systemd:
+    name: "{{ pp_servicename }}"
```
tanderegg, line 43: @rioh, see my comment on the other PR, I didn't think this was necessary, but let me know if you determined otherwise!

tanderegg (changes requested):

deploy/roles/shared/handlers/main.yml
rioh: @tanderegg apparenlty that option is only available with Ansible 2.4 and I think some other modules don't work with 2.4 currently.  Should we leave it on for now?

tanderegg: Ah, interesting.  OK, leave it as is for now, I will merge, thanks!

tanderegg (approved):

tanderegg:
@rioh It seems there's something up with the Travis test for db_server, but that was preexistant to this PR so I will just merge, we can investigate that later.

PR #311, opened 2018-01-05
Added go_server_url variable to gocd_server role

rioh:


deploy/roles/gocd_server/templates/go-server.j2
```diff
@@ -1,5 +1,6 @@
 export GO_SERVER_PORT={{ gocd_server_port }}
 export GO_SERVER_SSL_PORT={{ gocd_server_ssl_port }}
+export GO_SERVER_URL={{ go_server_url }}
```
tanderegg, line 3: @rioh Is this now needed by the Go Server itself?  Just want to verify, since in the past it was only needed on the agents to let them know how to contact the server.

rioh: @tanderegg no, in a previous build, Jenkins was complaining about the missing env variable, but after the other recent changes, it's no longer complaining. Maybe it was just a misleading error message.

PR #318, opened 2018-01-18
deploy RStudio Pro Drivers

stirlingv:


deploy/roles/r-studio-server/tasks/main.yml
```diff
@@ -26,6 +26,7 @@
     state: "{{ rstudio_server_package_state }}"
     disable_gpg_check: yes
     validate_certs: no
+  when: upgrade_rstudio
```
tanderegg: @stirlingv I think we already have an `update_rstudio` variable, do we need a separate one?

deploy/roles/r-studio-server/tasks/main.yml
```diff
@@ -40,3 +41,46 @@
...
+
+- name: install R Studio Pro drivers
+  get_url:
+    url: "{{ rstudio_pro_drivers_url }}"
+    dest: "{{ rstudio_pro_drivers_folder }}"
+  when: install_pro_drivers
+  tags:
+    - rstudio-server
+
+- name: set R Studio Pro dirver installer to executable
```
tanderegg: `dirver` -> `driver`

deploy/roles/r-studio-server-config/defaults/main.yml
```diff
@@ -4,6 +4,7 @@ rstudio_port: 8787
 rstudio_address: 0.0.0.0
 rstudio_pro: False
 rstudio_session_timeout: 15
+upgrade_rstudio: False
```
tanderegg: See comment below, `update_rstudio` already exists.

tanderegg (changes requested):

deploy/roles/r-studio-server/tasks/main.yml
stirlingv: @tanderegg I think that variable is for something different. I left `update_rstudio=false` when running the upgrade. If you don't think it matters, I'll happily modify to all use the same variable.

deploy/roles/r-studio-server/tasks/main.yml
stirlingv: d'oh!

deploy/roles/r-studio-server/tasks/main.yml
tanderegg: Ah, technically `update_rstudio` should be `True` when upgrading.  This ensures that all users are disconnected before the upgrade happens, which prevents some issues from occurring.

stirlingv: oh, I get it! sorry about that. I'll make the changes.

deploy/roles/r-studio-server/defaults/main.yml
```diff
@@ -1,3 +1,8 @@
 ---
 rstudio_server_package_name: https://download2.rstudio.org/rstudio-server-rhel-1.1.383-x86_64.rpm
 rstudio_server_package_state: present
+
+rstudio_pro_drivers_url: http://drivers.rstudio.org/7C152C12/odbc-install.sh
+rstudio_pro_drivers_folder: /etc/rstudio
```
tanderegg: @stirlingv /etc is meant for configuration files only, the installer should probably just go in /tmp

stirlingv:
@tanderegg I believe all these are addressed. I deployed to dev - I think the drivers are there; although, the only testing I did was to look at the tools in RStudio and notice that most of them are no longer greyed out.

deploy/roles/r-studio-server/defaults/main.yml
```diff
@@ -1,3 +1,8 @@
 ---
 rstudio_server_package_name: http://download2.rstudio.org/rstudio-server-rhel-1.1.383-x86_64.rpm
 rstudio_server_package_state: present
+
+rstudio_pro_drivers_url: http://drivers.rstudio.org/7C152C12/odbc-install.sh
+rstudio_pro_drivers_folder: /etc/rstudio
```
tanderegg: @stirlingv This should just be /tmp, since /etc is meant for permanent configuration files.

deploy/roles/r-studio-server/defaults/main.yml
```diff
@@ -1,3 +1,8 @@
...
+rstudio_pro_driver_install_name: odbc-install.sh
+rstudio_pro_driver_install_folder: /usr/lib
```
tanderegg: @stirlingv This is where I was thinking /opt/rstudio-server might make sense, perhaps /opt/rstudio-server/drivers?  You would also need a task to ensure that this directory exists and has the right permissions.

tanderegg (changes requested):

deploy/roles/r-studio-server/tasks/main.yml
```diff
@@ -40,3 +41,52 @@
...
+    executable: /usr/bin/pip
+  when: install_pro_drivers
+  tags:
+    - rstudio-server
+
+- name: Create RStudioPro driver folder
+  file:
+    path: "{{ rstudio_pro_driver_install_folder }}"
+    state: directory
+    mode: a+rwx
```
tanderegg: @stirlingv we just need `a+rx`, no `w` since we don't want folks changing anything!

tanderegg (changes requested):
One small change, otherwise we are good!

stirlingv:
@tanderegg This should be good to go now!

tanderegg (approved):
Thanks, looks good!

PR #319, opened 2018-01-18
Updating odbcinst file to include rstudio pro driver configurations

stirlingv:


deploy/roles/odbc/templates/odbcinst.ini
```diff
@@ -27,3 +27,27 @@ Setup                   = /usr/lib64/libtdsS.so
...
+Driver = /home/deploy/postgresql/bin/lib/libpostgresqlodbc_sb64.so
+
+[Salesforce]
+Driver = /home/deploy/salesforce/bin/lib/libsalesforceodbc_sb64.so
+
+[SQLServer]
+Driver = /home/deploy/sqlserver/bin/lib/libsqlserverodbc_sb64.so
+
+[Teradata]
+Driver = /home/deploy/teradata/bin/lib/tdataodbc_sb64.so
```
tanderegg: @stirlingv We shouldn't hardcode these in here, because they are only needed when the Pro drivers are installed.  So you shoul add a new variable called `odbcinst_entries`, that is a list of dictionaries:
```
odbcinst_entries:
  - driver_name: "AmazonRedshift"
    attributes:
        Driver: /home/deploy/amazonredshift/bin/lib/liubamazonredshiftodbc_sb64.so
```
Then, in the odbcinst.ini template, you could do, after the other entries:
```
[{{ driver.driver_name }}]
{% for key, value in driver.attributes.iteritems() %}
{{ key }} = {{ value }}
{% endfor %}
```
The above may not be 100% correct, but that's the general idea.

Also, it would be good if the drivers could be somewhere other than /home/deploy, ideally /usr/lib or somewhere more general.  Other users would not normally have access to the `deploy` user's folder.

tanderegg (changes requested):

tanderegg:
@stirlingv Can you take a look here too?

stirlingv:
For some reason it shows way more files changed than were changed...

PR #320, opened 2018-01-18
Test Citus Scripts

lel99999:
test citus scripts updated, renamed and moved to test folder


tanderegg:
@lel99999 Looks like this PR includes all the R changes too, would you mind removing those?  Thanks!

test/test_citusdb_baseRI.yml
```diff
@@ -0,0 +1,165 @@
...
+        dest: /tmp/events.csv
+        mode: 0755
+
+    - name: "create table users"
+      become: yes
+      become_user: postgres
+      become_method: sudo
+      command: >
+        psql -c "create table github_users(user_id bigint,url text,login text,avatar_url text,
+        gravatar_id text,display_login text);"
```
tanderegg: @lel99999 `become_user` doesn't work well with our infrastructure for some reason, so it is best to avoid it.  The alternative would be to use the `postgresql_exec` module I created, which allows you to pass `pp_superadmin` and `pp_superpassword` as parameters, and not use `psql`.  Please update this one and all those below.

test/test_citusdb_baseRI.yml
```diff
@@ -0,0 +1,165 @@
...
+      become_user: postgres
+      become_method: sudo
+      command: >
+        psql -c "SELECT count(*) FROM github_users"
+      register: select_counts_userdata_sql_results
+
+    - name: printout select counts on users results
+      debug:
+        msg: "{{ item[2] }}"
+      with_nested: "{{ select_counts_userdata_sql_results.stdout_lines }}"
```
tanderegg: @lel99999 You could potentially use the `assert` module instead of (or in addition to) the `debug` module, which would return an error if the count doesn't match the correct value: http://docs.ansible.com/ansible/latest/assert_module.html

test/test_citusdb_baseRIwPEmodule.yml
```diff
@@ -0,0 +1,259 @@
+---
+- hosts: citus_coordinator
```
tanderegg: @lel99999 We should probably set this up to run from a separate server to fully test connectivity. In Prod that would be wdcddtl01 or wdcddtl02, but we should probably have this controlled by a variable, such as hosts: {{ citus_test_client }} or something.  The CSV's must still be downloaded on the Coordinator for the COPY command to work properly without `psql`.

test/test_citusdb_baseRIwPEmodule.yml
```diff
@@ -0,0 +1,259 @@
...
+    - name: print sql results
+      debug:
+        msg: "{{ create_table_users_sql_results['status'] }}"
+        #msg="-----------------------------------------------" 
+      #with_nested: "{{ query_result.stdout }}"
+ 
+#   - name: Assert results are correct
+#     assert:
+#       that: 
+#         - "{{ create_table_users_sql_results['status'] }} s defined"
```
tanderegg: @lel99999 I would like to try to use `assert` if possible, for each test, along with the `debug`'s, since the goal is to have the Jenkins deployment job fail if these tests don't pass.

test/test_citusdb_baseRIwPEmodule.yml
```diff
@@ -0,0 +1,259 @@
...
+  tasks:
+    - name: Install requirements for postgresql Ansible tasks
+      yum:
+        name: "{{ item }}"
+        state: present
+      with_items:
+        - python-psycopg2
+
+    - name: Download users data (31.7MB)
+#     shell: wget https://examples.citusdata.com/tutorial/users.csv -O /tmp/users.csv
```
tanderegg: @lel99999 Please remove commented out lines that are no longer needed.

tanderegg (changes requested):
@lel99999 Thanks for cleaning this up, looks much better!  One high level thing, it seems that we only need the second of the two playbooks, that uses the `postgresql_exec` module.  The first won't work in our environment, since using the `become_user` command is unreliable for various reasons.  So let's just use the second playbook, running from a user-defined server, to do the tests.  Then, perhaps rename the playbook to somethings simpler, just `test_citus.yml` would work fine.  Thanks much!

test/test_citusdb.yml
```diff
@@ -0,0 +1,256 @@
+---
+- hosts: citus
```
tanderegg: This is where I was thinking we should use the variable to determine which host this runs from.  If it is set to `citus`, then the code below will be run on every Citus machine (so 13 times in Production), when we only want it to run once.  I think if you do `hosts: {{ citus_client_host }}`, then we can pass `--extra_vars={'citus_client_host': 'wdcddtl01'}` or whatever to run it from a remote machine.

test/test_citusdb.yml
```diff
@@ -0,0 +1,256 @@
...
+  tasks:
+    - name: Install requirements for postgresql Ansible tasks
+      yum:
+        name: "{{ item }}"
+        state: present
+      with_items:
+        - python-psycopg2
+
+    - name: Download users data (31.7MB)
+#     shell: wget https://examples.citusdata.com/tutorial/users.csv -O /tmp/users.csv
```
tanderegg: Please remove commented out lines if they are no longer needed, otherwise annotate them with some context.

test/test_citusdb.yml
```diff
@@ -0,0 +1,256 @@
...
+
+    - name: Download users data (31.7MB)
+#     shell: wget https://examples.citusdata.com/tutorial/users.csv -O /tmp/users.csv 
+      get_url:
+        url: https://examples.citusdata.com/tutorial/users.csv
+        dest: /tmp/users.csv
+        mode: 0755
+
+    - name: Download events data (74.7MB)
+#     shell: wget https://examples.citusdata.com/tutorial/events.csv -O /tmp/events.csv
```
tanderegg: here too

test/test_citusdb.yml
```diff
@@ -0,0 +1,256 @@
...
+      get_url:
+        url: https://examples.citusdata.com/tutorial/events.csv
+        dest: /tmp/events.csv
+        mode: 0755
+
+    - name: Ensure that a connection can be made
+      postgresql_exec:
+        login_user: "{{ pp_dba_username }}"
+        login_password: "{{ pp_dba_password }}"
+        login_host: citus_coordinator
```
tanderegg: This could very based on context, so we should also set this to a variable, such as `login_host: "{{ citus_server_host }}"'.  This way we can use the same script to test both the Prod and Staging clusters, but not both if we choose.

test/test_citusdb.yml
```diff
@@ -0,0 +1,256 @@
...
+      debug:
+        msg: "{{ sql_out }}"
+        #msg="-----------------------------------------------" 
+      #with_nested: "{{ query_result.stdout }}"
+
+    - name: create table users
+      postgresql_exec:
+        login_user: "{{ pp_superadmin }}"
+        login_password: "{{ pp_superpassword }}"
+        login_host: citus_coordinator
```
tanderegg: Use the `citus_server_host` for all of these.

test/test_citusdb.yml
```diff
@@ -0,0 +1,256 @@
...
+    - name: print sql results
+      debug:
+        msg: "{{ create_table_users_sql_results['status'] }}"
+        #msg="-----------------------------------------------" 
+      #with_nested: "{{ query_result.stdout }}"
+ 
+#   - name: Assert results are correct
+#     assert:
+#       that: 
+#         - "{{ create_table_users_sql_results['status'] }} s defined"
```
tanderegg: Remove commented out lines

test/test_citusdb.yml
```diff
@@ -0,0 +1,256 @@
...
+    - name: create distributed table users
+      postgresql_exec:
+        login_user: "{{ pp_superadmin }}"
+        login_password: "{{ pp_superpassword }}"
+        login_host: citus_coordinator
+        port: "{{ pp_serverport }}"
+        db: "postgres"
+        script: "SELECT create_distributed_table('github_users', 'user_id');"
+      register:  create_dist_table_users_sql_results
+      ignore_errors: true
```
tanderegg: I think we don't want to ignore errors here, right?  Same with the next two.

test/test_citusdb.yml
```diff
@@ -0,0 +1,256 @@
...
+
+
+    - name: import users data
+      postgresql_exec:
+        login_user: "{{ pp_superadmin }}"
+        login_password: "{{ pp_superpassword }}"
+        login_host: citus_coordinator
+        port: "{{ pp_serverport }}"
+        db: "postgres"
+        script: "copy github_users from '/tmp/users.csv\' with csv;"
```
tanderegg: Escape should not be necessary, as it is single quotes within double quotes.

test/test_citusdb.yml
```diff
@@ -0,0 +1,256 @@
...
+        #script: "\\copy github_users from '/tmp/users.csv' with csv;"
```
tanderegg: Remove commented out line

tanderegg (changes requested):

tanderegg (approved):

PR #321, opened 2018-01-19
Added postgresql-libs to old_packages list

rioh:


tanderegg:
@rioh Oops, didn't realize this was a merge to Master.  I try to have everything go to Develop first, then merge Develop into Master (unlike aurora-internal, which is two logically separate code bases).

PR #323, opened 2018-01-24
Added task to create symlink and removed hardcoded path from sqitch

rioh:


deploy/roles/sqitch/tasks/main.yml
```diff
@@ -40,7 +40,7 @@
   register: timezone
   changed_when: '"is up to date" not in timezone.stdout'
   environment:
-    PATH: "{{ postgres_install_dir|default('/usr/pgsql-9.4') }}/bin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin"
+    PATH: "{{ postgres_install_dir|default('/usr/pgsql') }}/bin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin"
```
tanderegg: Since the above change means that the symlink always points to "postgres_install_dir", it is probably fine to simplify this and just use '/usr/pgsql' only (below as well)

tanderegg (changes requested):

deploy/roles/sqitch/tasks/main.yml
```diff
@@ -50,7 +50,7 @@
   register: sqitch
   changed_when: '"is up to date" not in sqitch.stdout'
   environment:
-    PATH: "{{ postgres_install_dir|default('/usr/pgsql-9.4') }}/bin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin"
+    PATH: "{{ postgres_install_dir }}/bin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin"
```
tanderegg: @rioh Sorry, I was unclear, I meant do the opposite, just hardcode '/usr/pgsql' as the value, since it will always point to `postgres_install_dir` anyways :)

tanderegg (changes requested):

deploy/roles/sqitch/tasks/main.yml
rioh: Oops,  I see.

tanderegg (approved):

PR #329, opened 2018-01-31
Citus 72

rioh:


deploy/deploy_citus_server.yml
```diff
@@ -74,6 +74,17 @@
...
+      postgresql_exec:
+        login_host: localhost
+        login_user: "{{ pp_superaccount }}"
+        login_password: "{{ pp_superpassword }}"
+        port: "{{ pp_serverport }}"
+        db: "{{ item }}"
+        script: ALTER EXTENSION citus UPDATE
+      check_mode: no
+      with_items:
+        - postgres
```
tanderegg: We should probably just run this on all DB's that have the Citus extension installed.  Unfortunately, it doesn't seem like there is a command to show all databases with a given extension installed, so it may be easiest just to run the command on all databases, and ignore the error when it says `extension "citus" does not exist`.

tanderegg (changes requested):

tanderegg:
@rioh Would you mind rebasing this to cleanup the commits a bit?  Thanks!

PR #330, opened 2018-02-01
Fixed nginx, ruby, postgresql_exec and revolution-r travis errors

rioh:
I still don't understand why we need 'become: True' for the postgres handler, but we get permissions error otherwise.


deploy/library/postgresql_exec.py
```diff
@@ -169,7 +169,8 @@ def main():
     kw['status'] = cursor.statusmessage
 
     try:
-        kw['query_result'] = cursor.fetchall()
+        # cursor.fetchall returns a list of DictRow classes which breaks with ansible 2.4
+        kw['query_result'] = [list(row) for row in cursor.fetchall()]
```
tanderegg, line 173: @rioh Excellent, thanks!

deploy/roles/nginx/tasks/main.yml
```diff
@@ -13,6 +13,12 @@
...
 
+- name: Create Nginx group, if needed
+  group:
+    name: "{{ nginx_user }}"
+  tags:
+    - nginx
+
 - name: Create Nginx user, if needed
   user:
     name: "{{ nginx_user }}"
```
tanderegg, line 24: @rioh Should we explicitly set the `nginx_user`'s group to `nginx_group` here?

deploy/roles/ruby/tasks/main.yml
```diff
@@ -1,26 +1,28 @@
...
+  #     dest: /tmp/mpapis.asc
+  #   tags:
+  #     - ruby
+  # 
+  # - name: Import key
+  #   command: "gpg --import /tmp/mpapis.asc"
+  #   register: import_key
+  #   changed_when: '"not changed" not in import_key.stderr'
+  #   tags:
+  #     - ruby
```
tanderegg: @rioh assuming this is no longer needed, should we just remove rather than comment out?

deploy/roles/shared/handlers/main.yml
```diff
@@ -32,6 +32,7 @@
   command: "restart shiny-server"
 
 - name: restart postgres
+  become: True
```
tanderegg, line 35: @rioh Yeah, this is weird, but doesn't hurt anything I suppose!

tanderegg (changes requested):
A few minor changes, otherwise looks good, thanks!

tanderegg (approved):

PR #333, opened 2018-02-07
Removed conditional clause for python-libs

rioh:


tanderegg:
@rioh Yes, sorry I should have caught that!  It should always be run regardless of environment.

PR #335, opened 2018-02-08
Remove duplicate entries from PATH

mattboehm:
I noticed that my $PATH contains 320 entries, with only 21 unique entries.

This fix remedies some of that duplication. As the path is searched from left to right, removing the duplicated elements from the end should have no effect on the order that the path is searched.


tanderegg:
@mattboehm Yeah, thanks for tackling this, every time I look at PATH I cringe :)

PR #336, opened 2018-02-13
Pg10 upgrade

rioh:


deploy/roles/postgresql-server/vars/edb.yml
```diff
@@ -4,7 +4,8 @@
 # Installer URL and executable
 pp_installer_version: 9.6.2.7
 pp_download_url: "http://get.enterprisedb.com/ga/edb-as96-meta-9.6.2.7-2-linux-x64.tar.gz"
```
tanderegg, line 6: @rioh Should these two be updated to 10.x as well?

deploy/roles/postgresql-server/vars/edb.yml
```diff
@@ -4,7 +4,8 @@
...
-pp_install_path: as9.6
+pp_install_path: as10
+pp_old_install_path: as9.6
 pp_config_template: "postgresql.conf.95.j2"
 pp_bindir: /opt/PostgresPlus/9.6AS/bin
 pp_old_bindir: /opt/PostgresPlus/9.4AS/bin
```
tanderegg, line 11: @rioh Also here, > 10.0 and 9.6 respectively

deploy/roles/python-libs/defaults/main.yml
```diff
@@ -10,6 +10,6 @@ python3_packages:
   - cookiecutter
 
 python_pip_bin: /usr/local/bin/pip2.7
-python3_pip_bin: /usr/bin/pip3.4
+python3_pip_bin: /usr/local/bin/pip3.4
```
tanderegg: I think now that we are using the EPEL version of python3, it should be /usr/bin/ not /usr/local/bin, right?

deploy/upgrade_db_server.yml
```diff
@@ -72,48 +174,73 @@
...
+        login_host: localhost
+        db: postgres
+        script: "{{ set_dependency_scripts.script }}"
+        port: "{{ pp_serverport }}"
+      when: '"citus_coordinator" in group_names'
 
-- hosts: db_server
+- hosts: all_db
   become: yes
   become_method: sudo
```
tanderegg, line 247: @rioh Add `serial: 1` and `max_failure_percentage: 1%` as described here: http://docs.ansible.com/ansible/latest/playbooks_delegation.html#rolling-update-batch-size, and here: http://docs.ansible.com/ansible/latest/playbooks_delegation.html#maximum-failure-percentage, to ensure failure of any one host will stop the process.

tanderegg (changes requested):

deploy/roles/python-libs/defaults/main.yml
rioh: I'll take this out - wasn't supposed to be checked in.

tanderegg (approved):

deploy/upgrade_db_server.yml
tanderegg: @rioh Can you add the serial part here too?  This role will restart the server(s), so we should do that in order rather than parallel..

tanderegg (changes requested):
One last change

PR #338, opened 2018-02-26
odbcinst file to include rstudio pro driver configurations

stirlingv:


stirlingv:
@tanderegg This should be good now/finally

deploy/roles/odbc/templates/odbcinst.ini
```diff
@@ -27,3 +27,11 @@ Setup                   = /usr/lib64/libtdsS.so
 FileUsage               = 2
 Threading               = 2
 useNTLMv2               = True
+
+[RStudioPro]
```
tanderegg: @stirlingv One small change, `[RStudioPro]` isn't needed here, since driver.drivername defines the header for each section.  In theory, we should update the full template to only use `odbcinst_entries`, and remove all the hard coded sections, then only include the R Studio Pro drivers when necessary.  However, we can merge this for now (once you've removed the unneeded heading), and tackle that later.

stirlingv: @tanderegg DONE! I knew it was unnecessary, but did it anyway. Hopefully we're good to go now!

tanderegg (approved):

PR #339, opened 2018-03-01
Fixed privilege escalation issues

rioh:


deploy/group_vars/all_db/main.yml
```diff
@@ -3,19 +3,19 @@
 
 backup_metadata_scripts:
   -
-    "CREATE TABLE public.pg_dist_partition AS SELECT * FROM pg_catalog.pg_dist_partition;"
+    "CREATE TABLE IF NOT EXISTS public.pg_dist_partition AS SELECT * FROM pg_catalog.pg_dist_partition;"
```
tanderegg, line 6: @risa Should we just drop the tables first?  The next time the upgrade is performed, if the tables already exist with old data, then these commands would not update the data and so the backup would be bad, right?

deploy/upgrade_db_server.yml
```diff
@@ -101,24 +92,17 @@
...
       command: >
-               {{ pp_prefix }}/{{ pp_install_path }}/bin/pg_upgrade
-                  --old-datadir {{ pp_old_datadir }}
-                  --new-datadir {{ pp_datadir }}
-                  --old-bindir {{ pp_prefix }}/{{ pp_old_install_path }}/bin
-                  --new-bindir {{ pp_prefix }}/{{ pp_install_path }}/bin
-                  --link
-                  -U {{ pp_serviceaccount }}
-                  --check
+          sudo su - {{ pp_serviceaccount }} -c "{{ pp_prefix }}/{{ pp_install_path }}/bin/pg_upgrade --old-datadir {{ pp_old_datadir }} --new-datadir {{ pp_datadir }} --old-bindir {{ pp_prefix }}/{{ pp_old_install_path }}/bin --new-bindir {{ pp_prefix }}/{{ pp_install_path }}/bin --link -U {{ pp_serviceaccount }} --check"
```
tanderegg: Minor quibble, but would you mind keeping the newlines?  It makes this much easier to read.

deploy/upgrade_db_server.yml
```diff
@@ -131,20 +115,34 @@
...
-      become_user: "{{ pp_serviceaccount }}"
       command: >
-              {{ pp_prefix }}/{{ pp_install_path }}/bin/pg_upgrade
-                --old-datadir {{ pp_old_datadir }}
-                --new-datadir {{ pp_datadir }}
-                --old-bindir {{ pp_prefix }}/{{ pp_old_install_path }}/bin
-                --new-bindir {{ pp_prefix }}/{{ pp_install_path }}/bin
-                --link
-                -U {{ pp_serviceaccount }}
+          sudo su - {{ pp_serviceaccount }} -c "{{ pp_prefix }}/{{ pp_install_path }}/bin/pg_upgrade --old-datadir {{ pp_old_datadir }} --new-datadir {{ pp_datadir }} --old-bindir {{ pp_prefix }}/{{ pp_old_install_path }}/bin --new-bindir {{ pp_prefix }}/{{ pp_install_path }}/bin --link -U {{ pp_serviceaccount }}"
```
tanderegg: @rioh Ditto

tanderegg (changes requested):

deploy/group_vars/all_db/main.yml
rioh: @tanderegg The temp tables are dropped later if the upgrade completes. I had to add this in because the script failed after it took backups but before they were dropped. When I tried to rerun the job, it would complain the tables already existed.

tanderegg: Ah, OK as long as they are dropped we are good, thanks!

tanderegg (approved):

PR #341, opened 2018-03-19
update MathLM mathpass file

stirlingv:


deploy/group_vars/all/main.yml
```diff
@@ -185,6 +185,10 @@ restricted_group_name: admin
 
 # Research Environment
 install_odbc: True
+math_lm_server_name: research_gauss
+math_lm_math_id: math_id
+math_lm_activation_key: key
+math_lm_password: password
```
tanderegg: @stirlingv These should go in `roles/math_lm/defaults/main.yml`, no?  Generally we only want to put stuff in "all" if it applies to multiple roles or servers.

deploy/roles/math_lm/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+- name: Verify mathpass file exists
+  stat: path="{{mathpass_location}}"
+  register: mathpass
+  when: update_math_lm_activation_key
+  tags: math_lm
+
+- name: Copy and rename existing mathpass file
+  command: mv {{mathpass_location}} {{mathpass_location}}_{{date}}
+  when: mathpass.stat.exists and update_math_lm_activation_key
+  tags: math_lm
```
tanderegg: No need to do all this, just use the `backup` option for the `template` command, which will copy the file with timestamp automatically: http://docs.ansible.com/ansible/latest/template_module.html

tanderegg (changes requested):

deploy/roles/math_lm/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+update_math_lm_activation_key: False
+mathpass_location: /usr/local/Wolfram/MathLM/mathpass
```
tanderegg: @stirlingv This is no longer used

PR #342, opened 2018-04-06
Mesos upgrade to 1.4

tanderegg:


tanderegg:
@rioh @stirlingv Could you take a look when you have a minute?  Thanks!

PR #345, opened 2018-04-19
SAS - new dependency and change config so odbc works w/postgres

stirlingv:


deploy/roles/sas/tasks/main.yml
```diff
@@ -1,9 +1,11 @@
 ---
 - name: Install yum dependencies
   yum:
-    name: libXext.i686
+    name: "{{ item }}"
     state: present
-  when: install_sas
+  with_items:
+    - libXext.i686
```
tanderegg: @stirlingv We should take another look at why this is necessary, as it doesn't make sense to install an i686 package on an x86_64 CPU...

deploy/roles/sas/tasks/main.yml
```diff
@@ -36,6 +38,24 @@
   tags:
     - sas
 
+- name: check sasenv script exists sas-console
+  stat:
+    path: "{{ sas_home }}/SASFoundation/9.4/bin/sasenv_local"
```
tanderegg: @stirlingv Shouldn't we use `sas_version` here instead of 9.4?

deploy/roles/sas/tasks/main.yml
```diff
@@ -36,6 +38,24 @@
...
+  register: setenv_file_console
+  tags:
+    - sas
+
+- name: Change the library path sas-console
+  lineinfile:
+    dest: "{{ sas_home }}/SASFoundation/9.4/bin/sasenv_local"
```
tanderegg: `sas_version` here too?

tanderegg (changes requested):
A few small issues, thanks for doing this quick!

deploy/roles/sas/tasks/main.yml
stirlingv: @tanderegg I think you're right. I removed dependency for .i686 library

tanderegg:
@stirlingv Can you manually remove the libXext.i686 package from the Dev SAS VM, so we can confirm that it is not needed?  Thanks!

deploy/roles/sas/tasks/main.yml
stirlingv: @tanderegg yup!

deploy/roles/sas/tasks/main.yml
stirlingv: @tanderegg double yup!

stirlingv:
@tanderegg removed libXext.i686 in dev. Was able to open SAS, no issues. I think we're good!

PR #348, opened 2018-04-27
Added changes for adding a citus standby node

rioh:


deploy/deploy_citus_server.yml
```diff
@@ -74,7 +72,31 @@
...
+  become_method: sudo
+  serial: 1
   roles:
     - role: citus_coordinator
+  tasks:
+    - name: Create back up of postgresql.conf
+      copy:
+        src: "{{ pp_datadir }}/postgresql.conf"
+        dest: "/tmp/postgresql.conf"
+        remote_src: yes
```
tanderegg, line 101: @rioh Out of curiosity, what is this for?  /tmp is cleared on every reboot, so shouldn't be used as a long-term backup of any kind.

deploy/group_vars/all/main.yml
```diff
@@ -222,6 +222,11 @@ deploy_user: vagrant
 pp_dba_username: vagrant
 pp_dba_password: vagrant
 
+# Replication user
+replication_user: replication
+replication_password: replication
+
+pp_datadir: /var/lib/pgsql/10/data
```
tanderegg: @rioh I'm not positive, but I believe it would be fine just to put these in the `postgresql_server` role defaults.  As long as a playbook is run with that role included, any variables defined there should be available in other roles running within the same play.  In other words, the variables should only go in `group_vars/all` if they are used in multiple playbooks that don't necessarily include the same roles.  We have not been sticking to that so far, but we should go back and clean that up at some point.  I could be wrong though, so if you've already tried this and it didn't work, ignore this comment :)

deploy/group_vars/citus_coordinator/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+pp_wal_level: hot_standby
+pp_max_wal_senders: 5
+pp_wal_keep_segments: 32
+pp_archive_mode: on
+pp_archive_command: 'cp %p {{ pp_datadir }}/archives/%f'
+pp_max_replication_slots: 5
```
tanderegg, line 7: @rioh Just so you're aware, these variables will not be visible in Jenkins, since a different `group_vars` folder is used within `aurora_internal`.  So these settings will have to be fully provided in Prod.

deploy/roles/citus_coordinator_standby/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  become_user: "{{ pp_superaccount }}"
+  command: "/bin/pg_basebackup -h {{ hostvars['citus_coordinator']['ansible_ssh_host'] }} -U {{ replication_user }} -D {{ pp_datadir }}"
+
+- name: Update the postgres config file
+  become: True
+  become_user: "{{ pp_superaccount }}"
+  lineinfile:
+    dest: "{{ pp_datadir }}/postgresql.conf"
+    regexp: 'hot_standby = (.*)'
+    backrefs: yes
```
tanderegg: @rioh I don't think `backrefs` are necessary here, since the `line` part doesn't reference any.

deploy/roles/citus_coordinator_standby/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Move the data dir
+  become: True
+  become_user: "{{ pp_superaccount }}"
+  command: "mv /var/lib/pgsql/10/data /var/lib/pgsql/10/data.bk"
```
tanderegg: @rioh Should the standby's postgresql service be stopped before moving the data folder?

deploy/group_vars/citus_coordinator/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+pp_wal_level: hot_standby
+pp_max_wal_senders: 5
+pp_wal_keep_segments: 32
+pp_archive_mode: on
+pp_archive_command: 'cp %p {{ pp_datadir }}/archives/%f'
```
tanderegg, line 6: @rioh It doesn't look like `pp_archive_mode` and `pp_archive_command` are used anywhere else.

tanderegg (changes requested):
@rioh Mostly questions, let me know what you think!

deploy/deploy_citus_server.yml
rioh: The documentation stated to back up the file, I think mainly if something went wrong in the setup process, so we can easily revert back. I didn't think we needed it long term, so I put it in tmp. Should we keep it around?

deploy/group_vars/all/main.yml
rioh: you're right :)

deploy/group_vars/citus_coordinator/main.yml
rioh: ok.

deploy/roles/citus_coordinator_standby/tasks/main.yml
rioh: :)

deploy/roles/citus_coordinator_standby/tasks/main.yml
rioh: yup!  I had it, but must have deleted it when I was cleaning up the branch.  Thanks!

deploy/group_vars/citus_coordinator/main.yml
rioh: It's used in the postgres conf template.  Is that what you mean?

deploy/deploy_citus_server.yml
tanderegg: @rioh Got it.  Not sure if we need to do that, since the file is generated via Ansible anyways?  Not sure if there's anything sensitive in that file, but placing it in /tmp would make it readable by anyone, so we might want to remove this or put it somewhere else (or just set the permissions to 600), up to you!

deploy/group_vars/citus_coordinator/main.yml
tanderegg: Ah, sorry I had thought these were new settings, but you're just overriding them.  Makes sense!

deploy/group_vars/all/main.yml
tanderegg: @rioh Could you also remove these from here?  That way they are not set in multiple locations, which could lead to confusion down the road.

rioh: @tanderegg Turns out I did need those in group_vars.  I have a play within deploy_citus_server.yml that only has the postgresql-server-config role which don't have those variables defined.  Maybe add postgresql as a dependency for postgresql-server-config?

tanderegg: @rioh Actually, I think it makes sense to just copy them to postgresql-server-config.  Having postgresql-server as a dependency would cause it to run whenver postgresql-server-config is run, which isn't always desirable.

deploy/deploy_citus_server.yml
tanderegg: @rioh Could you just set the permissions to 600 for now, to be cautious?  Thanks!

tanderegg:
@rioh Last few comments, then I will merge and work on testing, thanks!

deploy/deploy_citus_server.yml
```diff
@@ -39,6 +39,21 @@
         src: "{{ pp_file_transfer_dir }}"
         state: link
 
+    - name: Check WAL directory exists
+      stat:
+        path: "{{ pp_xlogdir }}/pg_wal"
```
tanderegg: @rioh We should probably make "pg_wal" a variable, in case we need to deploy Postgres 9.x for some reason.

deploy/group_vars/citus_coordinator/main.yml
tanderegg: @rioh Just to confirm, do we want these settings in place for both coordinators?  If not should we have a separate group for the standby?

tanderegg (commented):
@rioh Just a few comments, thanks!

deploy/group_vars/citus_coordinator/main.yml
rioh: This is just for the master coordinator node and it doesn't get applied to the citus_coordinator_standby group.

tanderegg (approved):

PR #354, opened 2018-07-30
Updates for stata patch update

rioh:


deploy/roles/stata/tasks/main.yaml
```diff
@@ -73,12 +73,25 @@
   tags:
     - stata
 
+- name: Add the update do file
+  template:
+    src: stataupdate.do.j2
+    dest: "{{ stata_install_directory }}/stataupdate.do"
+  tags:
```
tanderegg: @rioh Should this also have `when: stata_patch_update`?

rioh: @tanderegg That was intentional, but I can add a when clause so it's obvious what it's for.

PR #355, opened 2018-08-01
add task to set r executable to research_sas

pratanaharris:
Update code to resolve this task  https://[GHE]/data/orca/issues/20.

### DIFF
--- a/deploy/deploy_research_environment.yml
+++ b/deploy/deploy_research_environment.yml
@@ -146,6 +146,16 @@
     - sas
     - gocd
     - stattransfer
+    - r
+
+  tasks:
+  - name: Set r executable to octl 444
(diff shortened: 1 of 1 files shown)


deploy/deploy_research_environment.yml
```diff
@@ -146,6 +146,16 @@
...
+    - r
+
+  tasks:
+  - name: Set r executable to octl 444
+    file:
+      path: '/usr/bin/R'
+      owner: root
+      group: root
+      mode: 0444
+    tags: run_r
```
tanderegg, line 158: @pratanaharris This will also prevent SAS from executing R, isn't that needed?  Or does SAS connect to R via libraries and never use the executable?

PR #360, opened 2018-08-24
Nofile fix

tanderegg:


tanderegg:
@stirlingv @rioh @lel99999 Could one of you take a look for errors?  Thanks!

PR #363, opened 2018-11-05
upgrade matlab to R2018b

stirlingv:


deploy/roles/matlab/defaults/main.yml
```diff
@@ -0,0 +1,29 @@
...
+
+
+matlab_installation_key:     installkey
+matlab_outputfile:           /usr/local/MATLAB/R2018b/output/mathworks.log
+matlab_license_location:     /usr/local/MATLAB/R2018b/licenses/license.lic
+matlab_network_license_file: /usr/local/MATLAB/R2018b/licenses/network.lic
+matlab_license_file:         proprietary_license_info
+matlab_license_server:       server
+matlab_license_daemon:       daemon
+matlab_license_symlink_src:  etc/lmboot
```
tanderegg: @stirlingv In addition to this symlink, there are two others: /etc/lmboot, and /etc/lmdown, that should be linked to $MATLAB_PATH/etc/lmboot and $MATLAB_PATH/etc/lmdown respectively.  These are used by the "flexnet" init.d script located at /etc/init.d/flexnet, which was created by Matlab installer and starts at system boot.  Since these links were not updated, the 2015 version of the license manager was running and folks couldn't start Matlab (due to patching).

tanderegg (changes requested):

stirlingv:
@tanderegg I think the latest commit should do it!

PR #367, opened 2019-04-18
changes to aurora to allow local testing of psql-authnz script

stirlingv:


tanderegg (approved):
@stirlingv Thanks, good work!

PR #368, opened 2019-05-03
Develop ancillaryrh7

lel99999:
CentOS6 ->  RHEL7.x

- umask                 ->    umask = 022 (normal user - default dir permissions 775, file permissions 664, root - 755/644) 
- devtools              ->    devtools-7
- open_office           -> 	  open_office 4.1.6
- evince                -> 	  evince 3 
- stattransfer          ->    stattransfer 14 (stlinux64_install)
- sublime               ->	  sublime 3 (build 34207 x86_64)
- texlive		            ->	  texlive (latest) 
- ultra-edit            ->    ultra-edit 18.0-0.40 (x86_64)

### Require free developer subscription:  https://developers.redhat.com/blog/2016/03/31/no-cost-rhel-developer-subscription-now-available/
1. sign-up for developer subscription
2. create file ~/RHACCOUNT which is referenced in deployment file to register VM on RHN and enable RHN Repo.  Add following to ~/RHACCOUNT:
 _uid: \<name\>@\<domain\>
_pwd: \<pwd\>


deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+    - name: make Sublime3 install directory
+      shell: mkdir -p /usr/local/sublime
+    - name: make stattransfer14 install directory
+      shell: mkdir -p /usr/local/StatTransfer14
+
+
+    - name: Extract repo to destination point
+      shell:  tar xvf /tmp/ancilRH7_repo.tar.gz -C /opt/ancillary_local
+    - name: Extract OpenOffice 4.1.6 
+      shell:  tar xvf /opt/ancillary_local/Apache_OpenOffice_4.1.6_Linux_x86-64_install-rpm_en-US.tar.gz -C /opt/ancillary_local/OO416
```
stirlingv, line 85: @lel99999 I think these tar files should be in the /artifacts folder on the jenkins machine

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
+- hosts: wdcreancil02
```
stirlingv, line 1: @lel99999 I think we want to merge this into the deploy_researcher_environment.yml file, and either add these tasks as roles in the research_ancillary_services hosts section, or make a new section for research_ancillary_services_rh7.

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+  become_method: sudo
+  pre_tasks:
+    - debug:
+        msg: "Variables: _uid = {{ _uid }}, _pwd = **********"
+#       msg: "Variables: _uid = {{ _uid }}, _pwd = {{ _pwd }}"
+    - name: register RHEL7.5
+      command: "/usr/bin/subscription-manager register --username {{ _uid }} --password {{ _pwd }} --auto-attach"
+      when: _uid != "" and _pwd != ""
+#     command: "/usr/bin/subscription-manager register --username <name> --password <password> --auto-attach"
+    - name: register RHEL7 repo
```
stirlingv, line 16: @lel99999 is this necessary in prod, or should we be using the satellite managed by SE?

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+      shell:  tar xvf /opt/ancillary_local/Apache_OpenOffice_4.1.6_Linux_x86-64_install-rpm_en-US.tar.gz -C /opt/ancillary_local/OO416 
+    - name: install bz2
+      yum:
+        name: bzip2
+
+    - name: Extract Sublime3 
+      shell:  tar xvf /opt/ancillary_local/sublime_text_3_build_3207_x64.tar.bz2 -C /opt/ancillary_local/Sublime3
+
+    - name: Extract Sublime3 to install dir
+      shell:  tar xvf /opt/ancillary_local/sublime_text_3_build_3207_x64.tar.bz2 -C /usr/local/sublime
```
stirlingv, line 94: @lel99999 should there be 2 extract tasks for sublime??

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+        path: /etc/yum.repos.d/local.repo
+        state: touch
+        mode: "u=rw,g=r,o=r" 
+    - name: edit local.repo file
+      shell: |
+        echo '[localrepo]
+        name=localrepo
+        baseurl=file:///opt/ancillary_local/OO416
+        enabled=1
+        gpgcheck=0' >> /etc/yum.repos.d/local.repo
```
stirlingv, line 114: @lel99999 What is the local repo for?

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+      shell: "/usr/bin/subscription-manager repos --list-enabled"
+      register: enabled_repos
+    - debug:
+        msg: "RHEL Enabled Repos: {{ enabled_repos }}" 
+# tasks:
+    - name: download epel
+      shell:  wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm -O /tmp/epel-release-latest-7.noarch.rpm   
+    - name: Install epel
+      shell:  yum -y localinstall /tmp/epel*rpm
+
```
stirlingv, line 32: Are these dependencies or part of the Ancillary services? Might be a good idea to group all dependencies that can be yum installed, and adding them in one task and using with_items:

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+    - name: Install python-pip
+      shell: yum install -y python27 python-pip
+    - name: install gdown
+      shell: pip install gdown  
+#     yum:
+#       name: python27-python-pip
+#       state: latest
+    - name: Install python 3.6 and pip
+      shell: yum install -y python36 python36-pip
```
stirlingv, line 50: Shouldn't we be calling the python3 role for this, not installing python separately?

deploy/deploy_ancillaryRH7.yml
stirlingv: Also, shouldn't we be using the existing sublime role? or a new sublime_rhe7 one if the install is different/requires other dependencies for rh7?

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+      shell: mkdir -p /opt/ancillary_local/Sublime3
+    - name: make Sublime3 install directory
+      shell: mkdir -p /usr/local/sublime
+    - name: make stattransfer14 install directory
+      shell: mkdir -p /usr/local/StatTransfer14
+
+
+    - name: Extract repo to destination point
+      shell:  tar xvf /tmp/ancilRH7_repo.tar.gz -C /opt/ancillary_local
+    - name: Extract OpenOffice 4.1.6
```
stirlingv, line 84: Shouldn't we use the existing open_office role for this?

deploy/deploy_ancillaryRH7.dev_local
```diff
@@ -0,0 +1,144 @@
...
+    - open_officeRH7
+#   - evince
+#   See line #39 (evince)
+#   - stattransfer
+    - stattransfer14
+#   - sublime
+    - sublime3
+#   - ultra-edit
+    - ultra-editRH7
+    - texliveRH7
```
stirlingv, line 131: @lel99999 Why are we calling these roles for the local install, but not the prod deployment? Looks like in prod we're deploying each role as a task in the playbook. I think we prefer to break them into roles. 

Also, why are there two separate playbooks for local and prod - typically we've used the when: custom_repo tag to be a determination between prod and local development.

deploy/deploy_ancillaryRH7.dev_local
```diff
@@ -0,0 +1,144 @@
...
+
+  post_tasks:
+#   - name: Install texlive
```
stirlingv, line 134: @lel99999 probably want to remove commented out code throughout.

deploy/Vagrantfile
```diff
@@ -352,6 +352,26 @@ Vagrant.configure(2) do |config|
     server.vm.network :private_network, ip: "10.0.1.18"
   end
 
+  # Research ancillary server1 on RedHat 7.5 and above
+  config.vm.define :research_ancillary1_services do |server|
+#   server.vm.box = "bento/centos-6.10"
```
tanderegg, line 357: @lel99999 Could you remove commented out lines?  Also, to stick with previous conventions please call this config `research_ancillary_services_7`

deploy/Vagrantfile
```diff
@@ -352,6 +352,26 @@ Vagrant.configure(2) do |config|
...
+    server.vm.box = "clouddood/RH7.5_baserepo"
+    server.vm.host_name = "research-ancillary01.test.dev"
+
+    server.ssh.forward_agent = true
+
+    server.vm.provision "ansible" do |ansible|
+      ansible.playbook = "deploy_ancillaryRH7.dev_local"
```
tanderegg, line 364: @lel99999 We should not have separate playbooks for local and production, the same playbooks should be used and controlled by variables that can be overwritten.

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+  vars:  
+    custom_repo: true
+  vars_files:
+    - ~/RHACCOUNT
+  become: yes
+  become_method: sudo
+  pre_tasks:
+    - debug:
+        msg: "Variables: _uid = {{ _uid }}, _pwd = **********"
+#       msg: "Variables: _uid = {{ _uid }}, _pwd = {{ _pwd }}"
```
tanderegg, line 11: @lel99999 Remove debug code

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
+- hosts: wdcreancil02
+  vars:  
+    custom_repo: true
```
tanderegg, line 3: @lel99999 We do not want to hard-code `custom_repo` to be true, it should only be true in remote environments.

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+    - name: download epel
+      shell:  wget https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm -O /tmp/epel-release-latest-7.noarch.rpm   
+    - name: Install epel
+      shell:  yum -y localinstall /tmp/epel*rpm
+
+    - name: install X Window System
+      shell: yum -y groupinstall "X Window System"
+
+    - name: install XFCE
+      shell: yum -y groupinstall "xfce"
```
tanderegg, line 37: @lel99999 I don't think we need to install `xfce` on the Ancillary, since the desktop lives on EOD and apps are connected to via X11 forwarding only.

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+
+    - name: install Evince
+      shell: yum -y install evince
+
+    - name: Install python-pip
+      shell: yum install -y python27 python-pip
```
tanderegg, line 43: @lel99999 There's already a role for python, we should use that instead.

deploy/deploy_ancillaryRH7.yml
tanderegg, line 50: @lel99999 Same here, there's a role for python3 that should be used.

deploy/deploy_ancillaryRH7.yml
```diff
@@ -0,0 +1,142 @@
...
+#       state: latest
+#    - name: Build custom local repo
+#      shell: wget https://drive.google.com/open?id=1qaZH9_SrGyjHjWT3P2WtvZAhEkP9YG1Z /tmp
+    - name: make localrepo directory
+      shell: mkdir -p /opt/ancillary_local
+    - name: make OpenOffice directory
+      shell: mkdir -p /opt/ancillary_local/OO416
+    - name: make Sublime3 directory
+      shell: mkdir -p /opt/ancillary_local/Sublime3
+    - name: make Sublime3 install directory
```
tanderegg, line 76: @lel99999 What are these folders for?

deploy/deploy_ancillaryRH7.yml
tanderegg: @lel @stirlingv Yep, use existing roles when possible.

deploy/roles/stattransfer14/defaults/main.yaml
```diff
@@ -0,0 +1,13 @@
...
+stattransfer_local_bin_directory: /opt/ancillary_local
+stattransfer_license_type: "25 User Lease"
+stattransfer_expiration_date: "August 31, 2018"
+stattransfer_reg_code: STAT-TRANSFER-REG
+stattransfer_auth_code: STAT-TRANSFER-AUTH
+stattransfer_sign_code: STAT-TRANSFER-SIGN
+stattransfer_major_version: "14"
+stattransfer_name: "STAT-TRANSFER-NAME"
+stattransfer_organization: "STAT-TRANSFER-ORGANIZATION"
+stattransfer_set_code: "STAT-TRANSFER-SET-CODE"
```
tanderegg, line 13: @lel99999 Ideally we won't have a new role for stattransfer14 vs 13, just update the existing role with the new variable values.  That way we can control both versions using the same role while incrementally upgrading machines.

tanderegg (changes requested):
@lel99999 Left some comments in addition to @stirlingv's comments, can you review?  Thanks!

deploy/deploy_ancillaryRH7.yml
lel99999: the local repo duplicates the /artifacts functionality ... will upload there and mark as resolved.

deploy/deploy_ancillaryRH7.yml
lel99999: have added an epel7 role, so will replace this with that role

deploy/deploy_ancillaryRH7.yml
lel99999: this is a specific directive for RHEL7, it could be placed in a role, but we probably would have to make more fine-grained than python3, probably will have to spell out python3.6, etc.

deploy/deploy_ancillaryRH7.dev_local
lel99999: just a marker to work out the diffs and conditionals before a single playbook covering multiple scenarios are integrated

deploy/deploy_ancillaryRH7.yml
lel99999: folders are for custom local repo to stage installation files like /artifacts, will update to use /artifacts instead

deploy/roles/ultra-editRH7/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+ue_uninstall_list: 
+  - UltraEdit-16.1-0.22.x86_64
+  - UltraEdit-16.1-0.18.x86_64
+  - UltraEdit-3.3-0.12.x86_64
+  - UltraEdit-16.1-0.22.x86_64
+  - UltraEdit-15.1-0.8.x86_64
+
+ue_install_version: UltraEdit-18.0-0.40.el7.x86_64
+ue_install_rpm: UltraEdit-18.0-0.40.el7.x86_64.rpm
```
stirlingv, line 10: @lel99999 did you add this rpm to the SE repo for managed rpms? Typically we prefer to yum install that way, rather than from source ourselves.

deploy/roles/ultra-editRH7/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+    name: "{{ item }}" #UltraEdit.x86_64
+    state: removed
+    disable_gpg_check: yes
+  with_items: "{{ ue_uninstall_list }}"
+  when: custom_repo
+  tags:
+    - UltraEdit
+
+- name: Install Ultra Edit
+  shell:  yum localinstall -y {{ ue_install_dir }}/{{ ue_install_rpm }}
```
stirlingv, line 12: @lel99999 see comment above. We want to send SE the RPM so that can add it to RH satellite.

deploy/deploy_ancillaryRH7.dev_local
```diff
@@ -0,0 +1,144 @@
...
+      shell:  yum -y localinstall /tmp/epel*rpm
+
+    - name: install X Window System
+      shell: yum -y groupinstall "X Window System"
+
+    - name: install XFCE
+      shell: yum -y groupinstall "xfce"
+
+    - name: install Evince
+      shell: yum -y install evince
```
stirlingv, line 40: There is a whole role for this, that seems to require installing many more dependencies. I'm going to remove this and call the evince role.

deploy/deploy_ancillaryRH7.dev_local
```diff
@@ -0,0 +1,144 @@
...
+
+    - name: Install python-pip
+      shell: yum install -y python27 python-pip
+    - name: install gdown
+      shell: pip install gdown  
+#     yum:
+#       name: python27-python-pip
+#       state: latest
```
stirlingv, line 48: @lel99999 we should call the python27 role here.

PR #370, opened 2019-05-14
installing dependencies on citus hosts

stirlingv:
These dependencies were necessary to install on citus coordinator so that we could run the psql-authnz script

### DIFF
--- a/deploy/deploy_citus_server.yml
+++ b/deploy/deploy_citus_server.yml
@@ -74,6 +74,7 @@
   serial: 1
   roles:
     - postgresql-server-config
+    - psql-authnz
   post_tasks:
     - name: Create Citus extension
       postgresql_exec:
--- a/deploy/roles/psql-authnz/tasks/main.yml
+++ b/deploy/roles/psql-authnz/tasks/main.yml
@@ -1,0 +1,22 @@
+- name: Install git
+  yum:
+    name: git
+    state: present
+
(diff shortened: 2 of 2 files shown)


deploy/deploy_citus_server.yml
```diff
@@ -68,6 +68,27 @@
...
+        - virtualenv
+        - python-devel
+        - libevent-devel
+        - gcc
+        - gcc-c++
+        - kernel-devel
+        - libxslt-devel
+        - libffi-devel
+        - openssl-devel
+        - openldap-devel
```
tanderegg: @stirlingv Might it make sense to put this into a new role for `psql_authnz` that we can eventually use to automate that deployment?  That's what all these dependencies are for, no?  Thanks!

stirlingv: @tanderegg yup, that's what these are for. Makes a lot of cense to put it in a new role. I'll do that, rebase and force push to update this PR whilst keeping the commit history clean.

tanderegg: @stirlingv Thanks, let me know when ready!

stirlingv: @tanderegg I've added the dependencies to a psql-authnz role, but not the automation of the deployment of the script itself... I've created an issue so we remember to actually do it: https://github.com/cfpb/aurora/issues/371

PR #372, opened 2019-06-13
Rh python36

dvbalen:
- Removed python3.4
- Added python3.6 from SCL
- Made python-libs role more generic
- Make common role less glusterfs specific
- Removed some deprecation warnings
- Fixed Vagrant .vault-file workaround


deploy/Vagrantfile
```diff
@@ -9,7 +9,7 @@ Vagrant.configure(2) do |config|
   # First make sure that an .vault_password file exists, and if it doesn't
   # then create an empty file
   if not File.file?('.vault_password')
-    FileUtils.touch('.vault_password')
+    File.write('.vault_password','NOT A PASSWORD **autogenerated by Vagrant**')
```
tanderegg, line 12: @dvbalen Thanks! :+1:

deploy/roles/python36-scl/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+    - hosts: servers
+      roles:
+         - { role: python36-scl, set_python36_default_python: True }
+
+
+Author Information
+------------------
+
+CFPB
```
tanderegg, line 47: @dvbalen Double 👍 for this!

deploy/roles/python36-scl/meta/main.yml
```diff
@@ -0,0 +1,60 @@
+galaxy_info:
+  author: your name
+  description: your description
+  company: your company (optional)
```
tanderegg: @dvbalen Might as well fill this out since we've included the galaxy metadata?

deploy/roles/r-libs/tasks/main.yml
```diff
@@ -35,40 +35,62 @@
...
+
 - name: Install R packages
-  command: /usr/bin/Rscript --slave --no-save --no-restore-history -e "if (! ('{{ item }}' %in% installed.packages()[,'Package'])) { install.packages(pkgs='{{ item }}', INSTALL_opts=c('--no-test-load'),repos=c('http://cran.rstudio.com/'),dep=TRUE); print('Added'); } else { print('Already installed'); }"
+  shell: source scl_source enable devtoolset-2 && /usr/bin/Rscript --slave --no-save --no-restore-history -e "if (! ('{{ item }}' %in% installed.packages()[,'Package'])) { install.packages(pkgs='{{ item }}', INSTALL_opts=c('--no-test-load'),repos=c('http://cran.rstudio.com/'),dep=TRUE, upgrade_dependencies=FALSE); print('Added'); } else { print('Already installed'); }"
   with_items: "{{ r_packages }}"
   register: r_result
   failed_when: "r_result.rc != 0 or 'had non-zero exit status' in r_result.stderr"
   changed_when: "'Added' in r_result.stdout"
-  environment: "{{ r_env_vars }}"
+  #environment: "{{ r_env_vars }}"
```
tanderegg: If we're going this route, just remove the `environment` line completely rather than commenting it out.

deploy/roles/r-studio-server-config/tasks/main.yml
```diff
@@ -48,6 +51,7 @@
   template:
     src: rsession.conf.j2
     dest: /etc/rstudio/rsession.conf
+  when: update_rstudio
```
tanderegg: @dvbalen I think we only want the `when: update_rstudio` conditional for things that should only happen when R Studio is being upgraded to a newer version, no?  We may want to update the configuration files when not upgrading from time to time, to change settings.

tanderegg (changes requested):
A few minor changes, but overall looks great!  One thing, a few of the Travis CI tests failed due to the following: `Template source files must be utf-8 encoded`, which is a new error in 2.5 due to template source files that are not fully utf-8.  Looks like one of the template files for Intellij is actually a PNG file, which looks like a hack someone did in the past, but should be easy to fix.  There was also a failure in Elasticsearch due to SSL, that hopefully should be an easy fix.  R-libs update times out, so maybe turn on verbosity there to get more output.  Lastly, SAS and Matlab have broken paths, fix if easy, otherwise just ignore errors.

deploy/roles/python36-scl/meta/main.yml
dvbalen: CC0 License?

tanderegg: Yep, that's our standard.

tanderegg:
@dvbalen Well done, sorry this seemingly simple request ended up such a beast :)

PR #387, opened 2020-02-13
WIP: Citus sso

tanderegg:
Deployed to staging cluster, will merge after UAT.

### DIFF
--- a/deploy/deploy_citus_server.yml
+++ b/deploy/deploy_citus_server.yml
@@ -51,8 +51,9 @@
     - name: Ensure symlink exists for WAL directory
       file:
         dest: "{{ pp_datadir }}/{{ pg_wal }}"
-        src: "{{ pp_xlogdir }}{{ pg_wal }}"
+        src: "{{ pp_xlogdir }}/{{ pg_wal }}"
         state: link
+      when: pp_datadir != pp_xlogdir
 
     - name: Setup host file
       lineinfile:
--- a/deploy/deploy_citus_server.yml
+++ b/deploy/deploy_citus_server.yml
@@ -85,7 +85,9 @@
         port: "{{ pp_serverport }}"
         db: "{{ item }}"
         script: CREATE EXTENSION IF NOT EXISTS citus
+      register: result
       check_mode: no
+      changed_when: '"already exists" not in result.notices'
       with_items:
         - postgres
 
@@ -102,6 +104,7 @@
         dest: "/tmp/postgresql.conf"
         remote_src: yes
         mode: "0600"
+      changed_when: false
 
     - name: Create archive directory
       file:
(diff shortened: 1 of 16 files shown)


deploy/roles/postgresql-server/tasks/main.yml
```diff
@@ -4,7 +4,7 @@
 
 - name: Include variable for os
   include_vars: rhel_7.yml
-  when: "{{ ansible_distribution_major_version }} == 7"
+  when: ansible_distribution_major_version == 7
```
gpacix: This will fail, since the version number is a string; should be `== "7"`.  I'm not going to tell you how much of my life I wasted discovering that.

deploy/deploy_citus_server.yml
```diff
@@ -84,7 +85,9 @@
         port: "{{ pp_serverport }}"
         db: "{{ item }}"
         script: CREATE EXTENSION IF NOT EXISTS citus
+      register: result
       check_mode: no
+      changed_when: '"already exists" not in result.notices'
```
gpacix, line 90: Should we also check the return code (`result.rc`)?

deploy/roles/postgresql-server-config/tasks/main.yml
```diff
@@ -157,9 +169,10 @@
   check_mode: no
   register: result
   failed_when:
-    - 'result.module_stdout is defined and "extension \"citus\" does not exist" not in result.module_stdout'
-    - 'result.module_stderr is defined and "extension \"citus\" does not exist" not in result.module_stderr'
+    - '"extension \"citus\" does not exist" in result.notices'
```
gpacix: We can probably simplify the quotes to: `'extension "citus" does not exist' in result.notices` (omitting outermost quotes)

deploy/roles/postgresql-server/templates/pg_ident.conf.j2
```diff
@@ -0,0 +1,3 @@
+{% for ident in pp_ident_mappings %}
```
gpacix, line 3: Can we put some explanatory comments at the top of the file? e.g.:
`# This is a list of identity mappings for use with Kerberos; cf. some-url-or-other`
`# columns: Identity Type, Identity Match, Postgres Role`

deploy/roles/postgresql-server/tasks/main.yml
tanderegg: Ah, thanks much!

deploy/roles/postgresql-server-config/tasks/main.yml
tanderegg: The problem is that if a statement starts with quotes, Ansible expects the whole thing to be a string to pass internally to `eval`, so it fails unless the double quoting is used.

deploy/roles/postgresql-server/templates/pg_ident.conf.j2
tanderegg: Done, thanks!

deploy/deploy_citus_server.yml
tanderegg: Unfortunately, the return code doesn't contain useful information, it is always 0 if the query succeeded in running at all, regardless of whether the query actually did anything.

PR #388, opened 2020-03-11
Citus standby replication

gpacix:


deploy/roles/odbc/templates/odbcinst.ini
```diff
@@ -30,7 +30,7 @@ useNTLMv2               = True
 
 {% for driver in odbcinst_entries %}
 [{{ driver.drivername }}]
-{% for key, value in driver.attributes.iteritems() %}
+{% for key, value in driver.attributes.items() %}
```
dvbalen, line 33: {% for key, value in list( driver.attributes.items() ) %}

if you want it to work in both 2 and 3

deploy/roles/psql-authnz/tasks/main.yml
```diff
@@ -5,18 +5,7 @@
 
 - name: Install dependencies
   yum:
-    name: "{{item}}"
+    name: ['python-devel', 'libevent-devel', 'gcc', 'gcc-c++', 'kernel-devel', 'libxslt-devel', 'libffi-devel', 'openssl-devel', 'openldap-devel']
```
dvbalen: For these long lists an item per line would be more readable

.devcontainer/Makefile
```diff
@@ -0,0 +1,15 @@
+this_container=$(shell docker ps --filter "name=${CONTAINERNAME}" -aq)
+devcontainers=$(shell docker ps --filter "name=devcontainer" -aq | grep -v ${this_container})
+
+buildbase:
+	cd ${PROJECT_ROOT}/.devcontainer && docker build -f Dockerfile-base --rm --no-cache -t centos7-sshd .
+deploy:
+	ansible-playbook ${PROJECT_ROOT}/deploy/deploy_citus_server.yml -i ${PROJECT_ROOT}/deploy/vagrant_hosts \
+	--ask-pass --user=super --ask-sudo-pass
```
tanderegg, line 8: This is more of a big picture comment, and doesn't need to be addressed in this PR, but should we look into using the Docker provider for Vagrant rather than a Makefile?  Otherwise we will basically need to recreate the Vagrantfile as a Makefile, with build steps for each playbook.

.devcontainer/bashrc
```diff
@@ -0,0 +1,20 @@
+export WORKSPACE=${WORKSPACE:-/workspace}
+cat $WORKSPACE/.devcontainer/welcome.txt
+
+alias deploy='ansible-playbook $WORKSPACE/deploy/deploy_nginx_proxy_server.yml -i $WORKSPACE/deploy/vagrant_hosts --ask-pass --user=super --ask-sudo-pass'
```
tanderegg, line 4: Might help to document what these commands are for, for example why is the Nginx proxy server called out explicitly here?

.devcontainer/docker-compose.yml
```diff
@@ -0,0 +1,118 @@
...
+      # security_opt:
+      #   - seccomp:unconfined
+  
+      # Overrides default command so things don't shut down after the process ends.
+      command: sleep infinity 
+      networks:
+        ansible_nw:
+          ipv4_address: 10.0.1.253
+  #please note: the spacing needs to line up w/ the other yaml templates
+    citus_coordinator:
```
tanderegg, line 43: This file also seems like another approach to doing what Vagrant does.  It is it easy with docker-compose to only spin up a few of the listed containers, and restart them individually?  Again, it might be easier to rely on Vagrant/Docker rather than create a new workflow.

.travis.yml
```diff
@@ -7,167 +7,189 @@ env:
   - container_id=$(mktemp)
 
  matrix:
-#  proxy server
+# 1  proxy server
   - >
+    JOB_NAME="Proxy Server"
```
tanderegg, line 12: :+1: More clarity is always helpful!

archive-wal-file
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
tanderegg: I think this might have been included accidentally, only the template in the `citus_coordinator` role is needed, right?

deploy/deploy_admin_terminal.yml
```diff
@@ -28,5 +28,5 @@
         name: ansible
         state: present
         umask: "0022"
-      sudo: no
+      #sudo: no
```
tanderegg: Did the `sudo: no` cause problems?  In generally the goal was just to ensure that ansible installed correctly, and allow users to deploy from the Admin Terminal manually  if necessary (we rarely if ever do, however).

deploy/deploy_citus_server.yml
```diff
@@ -1,6 +1,13 @@
 # deploy_citus_server.yml
 ---
-- hosts: citus
+- name: Install common and optional packages, JDK, Postgresql, and Citus
```
tanderegg, line 3: :+1:

deploy/deploy_citus_server.yml
```diff
@@ -1,6 +1,13 @@
...
+  hosts: citus
+  # Avoid nasty surprises when someone does "--extra-vars pp_install=False", etc.:
+  pre_tasks:
+  - set_fact:
+      install_pp: "{{ pp_install | default(False) | bool }}"
```
tanderegg: Are these essentially to sanitize the value?  Is it possible to use `pp_install:` at the beginning rather than `install_pp`, or does that not work?  I'm asking only because having two different named variables that are meant to do the same thing could be confusing.

deploy/deploy_citus_server.yml
```diff
@@ -103,15 +119,104 @@
...
-- hosts: citus_coordinator_standby
+- name: Set up Citus Coordinator Standby
+  hosts: citus_coordinator_standby
+  pre_tasks:
+  - set_fact:
+      install_pp: "{{ pp_install | default(False) | bool }}"
   serial: 1
   roles:
-    - {role: citus_coordinator_standby, when: "{{ pp_install }}"}
+    - {role: citus_coordinator_standby, when: "pp_install | default(False) | bool"}
```
tanderegg, line 156: Should we use `install_pp` here rather than re-sanitizing?

deploy/deploy_citus_server.yml
```diff
@@ -103,15 +119,104 @@
...
+        # TODO: make this a variable
+        path: /var/lib/pgsql/.ssh/id_rsa
+        comment: "key for postgres@{{ ansible_hostname }}"
+        owner: postgres
+        group: postgres
+        mode: "u=rw,go="
+        state: present # this is the default anyway
+      register: coord_key
+  
+    - name: Fix owner of the generated pub key
```
tanderegg, line 190: Annoying that it doesn't use the right values for both files!

deploy/deploy_citus_server.yml
```diff
@@ -103,15 +119,104 @@
...
+      file:
+        path: /var/lib/pgsql/.ssh/id_rsa.pub
+        owner: postgres
+        group: postgres
+        mode: "u=rw,go=r"
+
+    - name: Add standby's hostkey to postgres@coordinator
+      lineinfile:
+        path: "~postgres/.ssh/known_hosts"
+        insertafter: "{{hostvars[groups['citus_coordinator_standby'][0]].ansible_ssh_host_key_ecdsa_public}}"
```
tanderegg: Minor nitpick, but generally I try to leave spaces between the curly braces and their content for readability.

deploy/deploy_citus_server.yml
```diff
@@ -103,15 +119,104 @@
...
+    - name: Create .ssh directory for user postgres
+      file:
+        state: directory
+        path: "/var/lib/pgsql/.ssh"
+        owner: postgres
+        group: postgres
+        mode: "u=rwx,go="
+
+    - name: Make postgres@standby trust public key of postgres@coordinator
+      ## delegate_to: central_server_name ## what's this?
```
tanderegg: Not sure what the question is here, `delegate_to` allows the Ansible command to run on a different server with reference to the host in question.

deploy/deploy_citus_server.yml
```diff
@@ -103,15 +119,104 @@
...
+      # TODO: add a "when: coord_key.changed
+      authorized_key:
+         user: "{{ pp_superaccount }}"
+         key: "{{ hostvars[groups['citus_coordinator'][0]].coord_key.public_key}}"
+         state: present
+         path: /var/lib/pgsql/.ssh/authorized_keys
+         comment: "postgres@{{ hostvars[groups['citus_coordinator'][0]].ansible_hostname}}"
```
tanderegg: This and `key` are missing a trailing space before the curly braces.

deploy/roles/citus_coordinator/templates/archive-wal-file.j2
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+# Generated at {{ ansible_date_time.iso8601 }} for host {{ inventory_hostname }}
+# Use like this (in /var/lib/pgsql/10/data/postgresql.conf):
+# archive_command = '/var/lib/pgsql/10/data/archive-wal-file %p %f'
```
tanderegg, line 4: Great script, thanks!  After this initial review, I'll try and do a run-through on Vagrant, but it looks good.

deploy/roles/citus_coordinator_standby/handlers/main.yml
```diff
@@ -1,7 +1,14 @@
 ---
+# copied this from aurora/deploy/roles/shared/handlers/rhel7.yml
```
tanderegg, line 2: In my other PR, I actually removed the `restart postgres` handler from the `citus` role, since it was unnecessary given that any playbook running that role is certainly using the `postgres_server` role as well, which already provides the handler.  This keeps things DRY'er.

deploy/roles/citus_coordinator_standby/tasks/main.yml
```diff
@@ -50,7 +62,7 @@
   become: True
   file:
     dest: "{{ pp_datadir }}/{{ pg_wal }}"
-    src: "{{ pp_xlogdir }}{{ pg_wal }}"
+    src: "{{ pp_xlogdir }}/{{ pg_wal }}"
     state: link
```
tanderegg, line 68: This should probably also have the `when: pp_datdir != pp_xlogdir` conditional.

deploy/roles/common/defaults/main.yml
```diff
@@ -3,7 +3,7 @@
 gluster_mount_options: "defaults,acl,_netdev"
 shared_links: []
 
-skip_glusterfs: false
+skip_glusterfs: true
```
tanderegg: Could you explain this a bit?  It's probably fine, but I want to make sure that we're not skipping GlusterFS when we don't need to.

deploy/roles/gocd/tasks/main.yaml
```diff
@@ -10,10 +10,10 @@
...
   tags: gocd
 
-- name: Copy local to remote
-  copy: 
-    src: "{{ gocd_agent_package_url }}"
-    dest: "{{ gocd_agent_package_path }}/{{ gocd_agent_package_name }}"
+- name: Download GoCD Agent RPM
+  shell: "curl -o {{ gocd_agent_package_path }}/{{ gocd_agent_package_name }} {{ gocd_agent_package_url }}"
+  args:
+    creates: "{{ gocd_agent_package_path }}/{{ gocd_agent_package_name }}"
```
tanderegg, line 16: Was this a bug workaround?  Normally I try to use the Ansible module rather than `shell` whenever possible.

deploy/roles/iptables/tasks/main.yml
```diff
@@ -79,5 +79,11 @@
 - name: Allow outgoing
   command: "iptables -P OUTPUT ACCEPT"
 
+# We need initscripts so we can call /sbin/service in the next task
+- name: Ensure initscripts is installed
```
tanderegg, line 83: This is RHEL version independent?

deploy/roles/log-courier/tasks/main.yml
```diff
@@ -12,12 +12,12 @@
 
 - name: Install log-courier
   yum:
-    name: "{{ item }}"
+    name: [
```
tanderegg, line 15: Normally I find using the `- item` syntax easier to read, was this change required for some reason?

deploy/roles/pem-client/tasks/main.yml
```diff
@@ -7,7 +7,7 @@
   when: install_pem
 
 - name: Ensure umask is correct
-  sudo: no
+  #sudo: no
```
tanderegg: This was here to make sure that the `deploy` users was running the umask command as itself, and not as root.

deploy/roles/postgresql-server/defaults/main.yml
```diff
@@ -211,7 +214,7 @@ pp_wal_sender_timeout: 60s
 pp_max_replication_slots: 0
 
 # master
-pp_syncronous_standby_names: ''
+pp_synchronous_standby_names: ''
```
tanderegg, line 215: Nice catch :)  We just need to double check that the misspelled version isn't set in `aurora-internal` somewhere.

deploy/roles/postgresql-server/tasks/main.yml
```diff
@@ -80,6 +85,7 @@
     owner: "{{ pp_serviceaccount }}"
     group: "{{ pp_serviceaccount }}"
     mode: 0700
+  when: ansible_distribution_major_version == "hell_froze_over" # TODO: figure out correct check
```
tanderegg: What's the issue here?

deploy/roles/postgresql-server/tasks/main.yml
```diff
@@ -229,14 +241,14 @@
 
 - name: Install requirements for postgresql Ansible tasks
   yum:
-    name: "{{ item }}"
+    name: [
```
tanderegg, line 240: Same question here as above re: list syntax

deploy/roles/shared/handlers/main.yml
```diff
@@ -4,7 +4,7 @@
   when: (ansible_distribution == 'CentOS' or ansible_distribution == 'Red Hat Enterprise Linux') and ansible_distribution_version == 6
 
 - include_tasks: rhel7.yml
-  when: (ansible_distribution == 'CentOS' or ansible_distribution == 'Red Hat Enterprise Linux') and ansible_distribution_version == 7
+  when: (ansible_distribution == 'CentOS' or ansible_distribution == 'Red Hat Enterprise Linux') and ansible_distribution_version == "7"
```
tanderegg: Normally we don't include the distribution check, although I suppose this is technically a more robust approach.  For now perhaps we should just change this to check the version in order to be consistent with other roles.

deploy/roles/sqitch/tasks/main.yml
```diff
@@ -12,7 +12,7 @@
     - sqitch
 
 - name: Ensure umask is correct
-  sudo: no
+  #sudo: no
```
tanderegg: Same here as above.  I think we actually created a separate `umask` role, so perhaps that should just be used?

deploy/roles/stata/tasks/main.yaml
```diff
@@ -157,6 +157,11 @@
   tags:
     - stata
 
+- name: DEBUG - see what is in stata_install_directory
```
tanderegg: FYI, the install directory only exists in our real environment, since we don't have a local installer for Stata to use due to licesning.

deploy/vagrant_hosts
```diff
@@ -2,32 +2,32 @@
 
 ; Shared Component
 [admin_terminal]
-admin_terminal ansible_ssh_host=10.0.1.2
+admin_terminal_0 ansible_ssh_host=10.0.1.2
```
tanderegg: :+1: but perhaps we should start with `_1` to be consistent with the other entries (elasticsearch, file_server)

tanderegg (changes requested):
@gpacix Thanks for all your work on this, overall looks great!  These are mostly comments and questions, only a few actual changes required I believe.  I will next run this locally just to make sure it all works on my machine, thanks again!

deploy/roles/psql-authnz/tasks/main.yml
gpacix: I've now got the entries on separate lines when there are 3 or more of them.

.devcontainer/Makefile
gpacix: That's a good point. I'll look into it.

.travis.yml
gpacix: Yeah; I don't yet use the variable, but it shows up in the Travis build overview and greatly aided my understanding of what was going on.

deploy/deploy_admin_terminal.yml
gpacix: This was a change I made very early; I'll try this without sudo (i.e. with `sudo: no`) and see if it works.

deploy/deploy_citus_server.yml
gpacix: I agree that this isn't the clearest way to do it. I'll improve it.

deploy/deploy_citus_server.yml
gpacix: Probably. I'll look at this as part of clarifying the `pp_install` stuff.

deploy/deploy_citus_server.yml
gpacix: Agreed.

deploy/roles/citus_coordinator_standby/tasks/main.yml
gpacix: I saw that in your PR; I'll make sure this one is similar.

deploy/roles/common/defaults/main.yml
gpacix: I'll look into this.

deploy/roles/gocd/tasks/main.yaml
gpacix: Yes; this is a workaround for the "Failed to validate the SSL certificate" errors we were getting; it's simpler than trying to use Python 2.7.  It (separating download from install) pops up a couple other places, too, for the same reason.

deploy/roles/iptables/tasks/main.yml
gpacix: I'm 99% certain it is, because I've run the `iptables` role dozens of times on both CentOS 6.x and CentOS 7.x.  Do we need more assurance than that?

deploy/roles/log-courier/tasks/main.yml
gpacix: Yes: Ansible gives a deprecation warning, and says the `item` syntax will not be supported in Ansible 2.10.  I took Daniel's advice to make it more readable by putting each package in the list on its own line.

deploy/roles/pem-client/tasks/main.yml
gpacix: I'll revert it and test that, then.

deploy/roles/postgresql-server/tasks/main.yml
gpacix: (See my previous reply about the deprecation warning.)

deploy/roles/shared/handlers/main.yml
gpacix: Probably. Maybe common or shared should do a check that it's RHEL or CentOS, and give a warning (or even error) if it's not.

deploy/vagrant_hosts
gpacix: This should be an easy fix: it's very rare that anything refers to a specific host; almost all references are by group.

deploy/deploy_admin_terminal.yml
gpacix: I've reinstated these and updated them to `become: no`.

deploy/deploy_citus_server.yml
gpacix: In every role or playbook this branch touches, I've made sure there's a space after `{{` and a space before `}}`.

deploy/deploy_citus_server.yml
gpacix: (see previous)

deploy/roles/citus_coordinator_standby/tasks/main.yml
gpacix: Added the `when` conditional.

deploy/roles/common/defaults/main.yml
gpacix: I accidentally committed that, and have now changed it back to `false`.  (It didn't affect any of the Travis build's jobs, since they all set `skip_glusterfs` to `true`, except for Job 3, which doesn't use GlusterFS at all anyway.)

deploy/roles/pem-client/tasks/main.yml
gpacix: Reinstated (see above reply for more detail).

deploy/roles/shared/handlers/main.yml
gpacix: Dropped the `ansible_distribution` clauses, changed `ansible_distribution_version` to `ansible_distribution_major_version`, and fixed the unquoted 6 a couple lines above.

deploy/roles/sqitch/tasks/main.yml
gpacix: I'm not clear on why we're changing the deploy user's `.bashrc` to set the `umask`.  I reinstated the `sudo: no` as `become: no` (see above).  It might be good to use the `umask` role, but that's a bigger change and I figured I'd go with the quicker fix right now.

deploy/roles/stata/tasks/main.yaml
gpacix: Thanks.  I've removed the DEBUG task.

deploy/vagrant_hosts
gpacix: Narrator: it actually **was** an easy fix.

archive-wal-file
gpacix: Yes; deleted this file at some point.

deploy/roles/postgresql-server/tasks/main.yml
gpacix: That was for developing; I've removed the "when" clause now.

tanderegg (approved):

PR #390, opened 2020-03-31
Minor fixes to matlab named-user license changes

gpacix:
Remove write permissions on mlm.opt;
fix iteritems --> items throughout;
fix compat-libstdc++-33 package name to work with CentOS 7 and 6

### DIFF
--- a/deploy/library/postgresql_exec.py
+++ b/deploy/library/postgresql_exec.py
@@ -135,7 +135,7 @@ def main():
         "port":"port",
         "db":"database"
     }
-    kw = dict( (params_map[k], v) for (k, v) in module.params.iteritems()
+    kw = dict( (params_map[k], v) for (k, v) in module.params.items()
               if k in params_map and v != "" )
 
     # If a login_unix_socket is specified, incorporate it here.
--- a/deploy/roles/odbc/templates/odbcinst.ini
+++ b/deploy/roles/odbc/templates/odbcinst.ini
@@ -30,7 +30,7 @@ useNTLMv2               = True
 
 {% for driver in odbcinst_entries %}
 [{{ driver.drivername }}]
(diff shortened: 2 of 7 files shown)


deploy/library/postgresql_exec.py
```diff
@@ -135,7 +135,7 @@ def main():
         "port":"port",
         "db":"database"
     }
-    kw = dict( (params_map[k], v) for (k, v) in module.params.iteritems()
+    kw = dict( (params_map[k], v) for (k, v) in module.params.items()
```
tanderegg, line 138: Was this to support Python 3?  I believe it is usually actually run on Python 2.x in our environment, and in that environment iteritems is more efficient, but if we need to change it to support 3, that's fine too.

deploy/roles/odbc/templates/odbcinst.ini
```diff
@@ -30,7 +30,7 @@ useNTLMv2               = True
 
 {% for driver in odbcinst_entries %}
 [{{ driver.drivername }}]
-{% for key, value in driver.attributes.iteritems() %}
+{% for key, value in driver.attributes.items() %}
```
tanderegg, line 33: See comment above

deploy/library/postgresql_exec.py
gpacix: Yes; at some point in the testing process, `ansible-playbook` or `jinja2` was running under Python 3 and complained about the odbc templates.  Unless the collections are very, very long, I doubt we'll even notice the efficiency difference.

deploy/roles/odbc/templates/odbcinst.ini
gpacix: Ditto.

tanderegg (approved):

PR #394, opened 2020-10-13
Added Devtools7 and Tesseract roles

redspart:
# Added
- Tesseract installing role for the GoCD agents
- Devtools7 role

### DIFF
--- a/deploy/roles/tesseract/tasks/main.yaml
+++ b/deploy/roles/tesseract/tasks/main.yaml
@@ -32,6 +32,27 @@
     - make
     - make install
 
+- name: Get autoconf-archive
+  get_url:
+    url: "{{ autoconf_archive_url }}"
+    dest: /tmp
+
+- name: Extract autoconf-archive
+  unarchive:
+    copy: no
+    src: /tmp/autoconf-archive-2019.01.06.tar.xz
+    dest: /tmp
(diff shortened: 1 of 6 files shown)


tanderegg (approved):
Looks good, let's do a test deploy in Dev against your fork, and then merge in.

deploy/roles/tesseract/tasks/main.yaml
```diff
@@ -0,0 +1,110 @@
...
+    - PKG_CONFIG_PATH: /usr/local/lib/pkgconfig
+  args:
+    chdir: /tmp/tesseract-4.0.0/
+  with_items:
+    - ./autogen.sh CC={{ CC }} CXX={{ CXX }}
+    - ./configure CC={{ CC }} CXX={{ CXX }} --prefix=/usr/local --with-extra-libraries=/usr/local/lib 
+    - make
+    - make install
+
+- name: "Empty out tessdata"
```
tanderegg: @redspart Is this part necessary?  I would assume the task below would update the directory via Git, but perhaps this is needed in case there are non-git files in there?

redspart: It has been removed. Originally, we were installing all the languages and not just targeting English.

tanderegg (approved):

